PR #1, opened 2014-01-04
Switch visibility support to  BOOST_SYMBOL_EXPORT. Refs #2114

jhunold:
Found while cleaning up my git svn mirror after the conversion. The patch switches Boost.Python to the common BOOST_SYMBOL_\*  macros from Boost.Config. This includes some minor cleanups for gcc which does not like visibility attributes on forward declarations. Testest with gcc and msvc-9.0

### DIFF
--- a/include/boost/python/detail/config.hpp
+++ b/include/boost/python/detail/config.hpp
@@ -64,31 +64,18 @@
 #endif
 
 #if defined(BOOST_PYTHON_DYNAMIC_LIB)
-
-#  if !defined(_WIN32) && !defined(__CYGWIN__)                                  \
-    && !defined(BOOST_PYTHON_USE_GCC_SYMBOL_VISIBILITY)                         \
-    && BOOST_WORKAROUND(__GNUC__, >= 3) && (__GNUC_MINOR__ >=5 || __GNUC__ > 3)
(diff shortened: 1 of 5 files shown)


include/boost/python/detail/config.hpp
```diff
@@ -64,43 +64,30 @@
...
 #  endif
-
 #endif
 
 #ifndef BOOST_PYTHON_DECL
 #  define BOOST_PYTHON_DECL
 #endif
 
 #ifndef BOOST_PYTHON_DECL_FORWARD
-#  define BOOST_PYTHON_DECL_FORWARD BOOST_PYTHON_DECL
```
lukaszforynski: why these (line 99 & 103) were also changed - if they're using BOOST_PYTHON_DECL before? It looks like it might break configurations that relied on those default settings.. IMO it seems unrelated to the intended change of 'switching' to BOOST_SYMBOL_EXPORT change.

include/boost/python/object/instance.hpp
```diff
@@ -11,7 +11,7 @@
 
 namespace boost { namespace python
 {
-  struct BOOST_PYTHON_DECL_FORWARD instance_holder;
```
lukaszforynski: this is the offending line that is a cause of #131

PR #2, opened 2014-02-08
[doc] indexing_suite and related classes are in boost/python/suite/indexing/

mlang:
Fix indexing.html to mention correct header paths.


stefanseefeld:
This patch is obsolete.

PR #5, opened 2014-03-18
Clean up master after subversion conversion.

danieljames:
This cherry picks some old commits from develop, removes a file from the CVS days and then sets up a sensible merge base from develop.


stefanseefeld:
What is the status of this pull request; is it still relevant ? It should be applied to the develop branch rather than master, at least. But I also see a few commits that appear obsolete, so I wonder whether we can't just close it..

stefanseefeld:
This is obsolete.

PR #6, opened 2014-03-20
Only use 'pragma once' when supported.

danieljames:
The checks were removed in 42e7d7bbb3c2b0994169b2d02b9e6767db90b08a. This is a little better as most compilers support it nowadays.


danieljames:
Sure, if that's what you prefer.

stefanseefeld:
Close as obsolete.

PR #7, opened 2014-04-22
Boost.Build now uses src/ instead of v2/

PeterMartini:
Update examples to reflect this


introom:
This definitely is a stumbling block for ppl never heard of boost.build before.

It's a easy and manifest fix, wonder why after a year's time, it's still un-merged.

stefanseefeld:
Thanks for the patch.

PR #10, opened 2014-09-09
override: check for NULL when calling overrides

mathstuf:
PyEval_CallFunction can return NULL in the event of an error which
causes method_result to error out. Instead, check for NULL and throw an
exception since Python has already set it.

### DIFF
--- a/include/boost/python/override.hpp
+++ b/include/boost/python/override.hpp
@@ -131,13 +131,17 @@ template <
 detail::method_result
 operator()( BOOST_PP_ENUM_BINARY_PARAMS_Z(1, N, A, const& a) ) const
 {
-    detail::method_result x(
-        PyEval_CallFunction(
-            this->ptr()
-          , const_cast<char*>("(" BOOST_PP_REPEAT_1ST(N, BOOST_PYTHON_FIXED, "O") ")")
-            BOOST_PP_REPEAT_1ST(N, BOOST_PYTHON_fast_arg_to_python_get, nil)
-        ));
-    return x;
+    PyObject* const result = PyEval_CallFunction(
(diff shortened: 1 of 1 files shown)


stefanseefeld:
Please add a failing test case (which passes after this patch is applied).
And please resubmit the pull request against the develop branch.
Thanks !

teeks99:
This should probably be merged into develop first, so that we can try it with the test runners before we put it in master. I'm not sure if that needs to happen in a different pull request.

stefanseefeld:
As suggested before, please submit this (with test case) agains the develop branch.

mathstuf:
There's no easy way to get this thing to return `NULL`. The only ways I can see are:

  - not passing in a tuple object (not possible to trigger from the API);
  - failing to allocate a tuple object (not easy to trigger);
  - kwargs not being a dictionary (not passed here);
  - internal recursion (not sure how exactly to test this);
  - returning `NULL` from a bound function while not setting an error (probably easiest).

So I've written the code which I think should show a failure…but how the heck do I get the test suite to actually run?

mathstuf:
Ping?

stefanseefeld:
As per my previous comment: Please submit a PR (including a test case) against the develop branch. Thanks !

mathstuf:
No, in response to my question as to how to run the test suite. I can't verify the test actually works on my machine.

stefanseefeld:
Oh, I see. Please have a look at my recent work on a SCons-based build system. Use
```
scons config
scons
scons test
```
to configure, build, and test respectively. Tests are in tests/, and are added to test/SConscript.

Please open a new PR where I can provide specific feedback to get this working. This (current) PR is closed.

PR #11, opened 2014-11-03
Hold the GIL while releasing shared_ptr_deleter's python handle.

jloy:
Python requires that the GIL be held while releasing objects, but requiring
that all clients of boost::shared_ptr-held types do so explicitly is
error-prone.  In the case where the wrapped type is consumed in a library that
you don't control, it might not even be possible to do correctly.  Instead,
have shared_ptr_deleter acquire the GIL on client's behalf at exactly the time
when it is necessary.

There's already an open trac issue for this (https://svn.boost.org/trac/boost/ticket/8290)
but the documentation for handle::release doesn't specify that it never throws, so the
patch in trac would fail to properly release the GIL state in that circumstance.

### DIFF
--- a/src/converter/builtin_converters.cpp
+++ b/src/converter/builtin_converters.cpp
@@ -30,6 +30,7 @@ shared_ptr_deleter::shared_ptr_deleter(handle<> owner)
 
 shared_ptr_deleter::~shared_ptr_deleter() {}
 
+#if PY_VERSION_HEX >= 0x02040000 && defined(WITH_THREAD)
 namespace
 {
 
@@ -37,23 +38,32 @@ namespace
   {
    public:
       scoped_gil_state()
-          : m_gil_state(PyGILState_Ensure())
-      {}
+      {
+          if ((m_py_threads_initialized = PyEval_ThreadsInitialized())) {
+              m_gil_state = PyGILState_Ensure();
+          }
+      }
 
       ~scoped_gil_state()
       {
-          PyGILState_Release(m_gil_state);
+          if (m_py_threads_initialized) {
+              PyGILState_Release(m_gil_state);
+          }
       }
 
    private:
       PyGILState_STATE m_gil_state;
+      bool m_py_threads_initialized;
   };
 
 }
+#endif
 
 void shared_ptr_deleter::operator()(void const*)
 {
+#if PY_VERSION_HEX >= 0x02040000 && defined(WITH_THREAD)
     scoped_gil_state gil;
+#endif
     owner.reset();
 }
 
--- a/src/converter/builtin_converters.cpp
+++ b/src/converter/builtin_converters.cpp
@@ -30,8 +30,30 @@ shared_ptr_deleter::shared_ptr_deleter(handle<> owner)
 
 shared_ptr_deleter::~shared_ptr_deleter() {}
 
+namespace
+{
+
+  class scoped_gil_state
+  {
+   public:
+      scoped_gil_state()
+          : m_gil_state(PyGILState_Ensure())
+      {}
+
+      ~scoped_gil_state()
+      {
+          PyGILState_Release(m_gil_state);
+      }
+
+   private:
+      PyGILState_STATE m_gil_state;
+  };
+
+}
+
 void shared_ptr_deleter::operator()(void const*)
 {
+    scoped_gil_state gil;
     owner.reset();
 }
 


eudoxos:
Great pull request. What is the status, is there a way to push it forward? 

For the record, link to old ticket: https://svn.boost.org/trac/boost/ticket/8290 and a discussion at c++-sig where some resistance was met.

eudoxos:
Ping?

stefanseefeld:
I'm aware of this PR. As has been pointed out earlier there were objections raised to this in http://code.activestate.com/lists/python-cplusplus-sig/17196/, as such a change may cause a deadlock.
Can you convince me that this isn't the case ? If not, could the code be reworked a bit to make this new policy optional, so users could pick explicitly whether they want to use a lock or not ?

eudoxos:
I see the objection as ungrounded; if there were potential deadlock with this change, there would be very likely crash with the current state - when Python object is deleted (or in general, when Python API is called), the acting thread *must* hold the GIL (I don't have citation now, but it is in Python docs written very clearly IIRC), otherwise other thread might simultaneously call Python API possibly causing crash, if both happen to touch the same memory structures.

If the behavior were selectable via policies, I'd need some guidance how to do it; or maybe someone else knowing that template jungle (meant in a good way) better could code it quickly.

nwhitehorn:
The only way to reliably avoid deadlocks here would be to have shared_ptr_deleter add reset requests to a global queue that is processed when the GIL is acquired in the normal course of events. This would require a special boost python wrapper for the GIL that knows to check the queue, but that isn't so bad.

eudoxos:
@nwhitehorn, I don't see how this would cause deadlock. Repeating, if GIL is in the same thread, the lock is a no-op, so no deadlock. If another thread tries to acquire it, it will wait till it is released. If it is already held in another thread, the request for GIL will be pending until the other thread releases it.

If there are cross-dependencies between threads leading to deadlock, then they would be crashing python as well (if not, only by accident).

Please reconsider this. I am now solving another bug due to this erratic design decision, and I am not very happy about it. (It might well be that boost::python is past its zenith in terms of being set in stone including its mistakes, for the fear of not breaking existing broken code and pybind11 is the rising star. Who knows...)

nwhitehorn:
The issue is that your code can't acquire any *other* locks while holding the holding the GIL, because the GIL release via shared_ptr deleter can happen at unpredictable times. This is an issue because all boost::python code is called with the GIL held unless explicitly released.

Suppose that you have a lock I will call A. The lock will be acquired in a C++ function called by boost::python I will call B(), but has no been yet -- though B() has started.

Separately, some other thread (C) is running and holds A. A shared_ptr held by C() goes out of scope while still holding A, which happens to have come from Python. The shared_ptr deleter then tries to acquire the GIL. That blocks, since B() is holding the GIL still.

Now B() tries the acquire lock A. That blocks on the completion of C() -- which is now held on the completion of B(), which is waiting for C(). And now you have a deadlock. Note that the only thing required for this to happen is that you have a lock -- any lock -- acquired by C++ code called into by boost python.

The problem arises in that shared_ptr deleters run in unpredictable contexts, so acquiring big locks is dangerous. Practically, one of the following needs to be true:
1. You can never let a shared_ptr go out of scope when any locks besides the GIL are held. In general, this is near-impossible, but the reset queue idea I mentioned actually does let this happen.
2. Make sure that any call into C++ from Python releases the GIL immediately, so that no C++ locks are ever acquired while holding the GIL. To make this reliable, this would need to get pushed into boost::python. All entry points into the interpreter then need to acquire the GIL, including explicit CPython calls not part of boost::python, which breaks the boost::python API.

lightmare:
I think that potential for deadlocks is not a valid argument against this PR.

In the example @nwhitehorn presented, whether `shared_ptr_deleter` acquires the GIL is secondary, the application is asking for trouble either way. As long as `shared_ptr_deleter` does not acquire the GIL, whoever released the GIL must make sure no shared pointers go out before it's re-acquired **(1)**. The thread C() dropped a pointer while not holding GIL -- `shared_ptr_deleter` now has a choice:
  - try to acquire the GIL and hope it won't deadlock, or
  - DECREF without the GIL and hope it won't crash.

Besides, trying to acquire any other lock while holding the GIL sounds like a bad strategy. It's a classic example where two threads are trying to acquire two locks, but in different order: thread B() goes GIL-->A, thread C() goes A-->GIL. The GIL is probably the most globalest lock in the application, so it should be acquired last.

**(1)** or more precisely, no shared pointers holding Python objects whose `tp_dealloc` is not thread-safe, e.g. containers (global trash can), unicode (global interned dict), float (global free list), ...

Reset queue sounds neat, might even be easy to implement with Py_AddPendingCall, but it seems a bit overkill compared to this PR.

PR #16, opened 2015-04-11
Fixes empty example

teeks99:
Uses of visual studio were getting a 'unresolved external symbol _mainCRTStartup' because there was no main() being linked into the test_embed.exe. I'm not a bjam expert, but I think that is because the embedding target is an executable (instead of an obj/lib?). I also have no idea if this was a problem on unix or if this fix will affect that platform.


teeks99:
I've now been able to run this on linux as well, and can verify that it doesn't cause any problems with the examples.

teeks99:
Any chance of getting this in before the freeze for 1.59 coming up next week?

PR #18, opened 2015-04-29
GCC warns about unused local typedef

alkino:
I only apply those patchs: https://svn.boost.org/trac/boost/ticket/8888

I don't own those patchs but I need them.

### DIFF
--- a/include/boost/python/args_fwd.hpp
+++ b/include/boost/python/args_fwd.hpp
@@ -36,15 +36,6 @@ namespace detail
       BOOST_STATIC_CONSTANT(std::size_t, size = 0);
       static keyword_range range() { return keyword_range(); }
   };
-
-  namespace error
-  {
-    template <int keywords, int function_args>
-    struct more_keywords_than_function_arguments
-    {
-        typedef char too_many_keywords[keywords > function_args ? -1 : 1];
-    };
-  }
 }
 
 }} // namespace boost::python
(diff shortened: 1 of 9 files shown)


teeks99:
I was just going to look at this and make sure it didn't upset anything in MSVC land, but I noticed that the pull request is against master instead of develop. I'm not the owner of this repo, but I thought all of boost was supposed to merge things to develop first. 

Maybe this pull request should be closed and a new one opened against develop instead, that would make it much easier for me to check out.

Tom

stefanseefeld:
Indeed. I'm not sure the pull request can be changed to point to the devel branch instead, or whether this requires a new pull request.

Aside from that, I'm not entirely sure I understand the effect of the patch (sorry, haven't had time to really look into it yet).
While I'm all for avoiding warnings, I'm somewhat afraid because it seems lots of the constructs here were to add error reporting capabilities, so the compiler would produce error messages that give the user a hint at what is missing (a poor man's substitute for concept checks).
Other than that, I think the patch(es) look good, and I would like to merge them (to devel).

blastrock:
I think a new pull request must be made, but surely there should not be conflicts rebasing this over the devel branch.

As for your concerns about the error reporting, this patch replaces the typedefs with BOOST_MPL_ASSERT_MSG which, I guess, just does the same thing (creating a typedef with the name provided, but with __attribute__((unused)) to avoid the warnings). Only the first occurrence has been replaced by BOOST_STATIC_ASSERT instead, but that can be easily fixed.

PR #33, opened 2015-08-12
Unset 'boost.defaults' set in Jamroot.

danieljames:
Not using '<format>html' as it seems it has to match the original
setting exactly.


stefanseefeld:
Thanks, merged !

PR #43, opened 2015-10-10
Fix a compile error with obsoleted details.

Flast:
Because boost/detail/iterator.hpp is now obsoleted, see boostorg/iterator@b2b9ab156881e700858c5a104c47e006eddcc717 .


Flast:
ping?

PR #44, opened 2015-10-14
Couple of Null check corrections

ya1gaurav:
1. m_to_python is checked against Null at line 70, but not checked at 89.
2. !! is same as without using this.

### DIFF
--- a/src/converter/registry.cpp
+++ b/src/converter/registry.cpp
@@ -85,10 +85,13 @@ BOOST_PYTHON_DECL PyObject* registration::to_python(void const volatile* source)
 
         throw_error_already_set();
     }
-        
-    return source == 0
-        ? incref(Py_None)
-        : this->m_to_python(const_cast<void*>(source));
+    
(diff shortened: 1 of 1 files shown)


ya1gaurav:
Please review updated patches. Thanks.

src/converter/registry.cpp
```diff
@@ -96,9 +98,13 @@ namespace
...
   void delete_node( T* node )
   {
-      if( !!node && !!node->next )
+      if( node && node->next )
+      {
           delete_node( node->next );
-      delete node;
+          delete node;
+      }
+      else if( node )
```
hotgloupi: You are checking if `node` is not null two times instead of one. And the `delete` operator is a no-op when the pointer is null, this whole thing can be simply written as:

    if (node != 0)
        delete_node(node->next);
    delete node;

ya1gaurav: I agree delete NULL is safe.

PR #47, opened 2015-11-03
Remove not reachable condition.

ya1gaurav:
Line no 138-139 suggest if condition satisfy if n_actual <=max_arity :
        if (n_actual + f->m_nkeyword_values >= min_arity
            && n_actual <= max_arity)
So condition at Line no 161 is not reachable.


hotgloupi:
LGTM

PR #50, opened 2015-12-23
Fix FAQ bottom link issue

SirChristian123:
Fix the FAQ bottom link dead issue that it suppose to lead to the "Fatal error C1204" documentation but leads to a 404 no page found error.


stefanseefeld:
This seems to be fixed now (http://boostorg.github.io/python/doc/html/faq/fatal_error_c1204_compiler_limit.html)

Issue #52, opened 2016-01-08
boost::python::file_exec not fully closing file with Python 3.4.3


asies:
Using Python 3.4.3 and Boost 1.57.0 on 64-bit Windows 7 with Visual Studio 2013.

After boost::python::exec_file is called on a file, that file can't be modified by the program using fopen_s, I think because it is still open.

In the following example, when ModifyFile is called before executing the file, it succeeds. After the file has been executed, fopen_s returns 13 (Permission denied) and ModifyFile fails.

```python
#include "stdafx.h"
#include "boost\python.hpp"
#include <iostream>

#define FILE_NAME         "myTest.py"

void ModifyFile( std::string newContent )
{
  FILE* myFile = nullptr;
  errno_t result = fopen_s( &myFile, FILE_NAME, "wb" );
  if ( result == 0 )
  {
    fwrite( newContent.c_str( ), sizeof( byte ), newContent.length( ), myFile );
    fclose( myFile );
    std::cout << "Success" << std::endl;
    return;
  }

  std::cout << "Failure" << std::endl;
}

int main( int argc, char** argv )
{
  Py_Initialize( );

  ModifyFile( "print(\"Hello\")" );

  boost::python::api::object mainNamespace = boost::python::import( "__main__" ).attr( "__dict__" );
  boost::python::exec_file( FILE_NAME, mainNamespace, mainNamespace );
  
  ModifyFile("print(\"Goodbye\")");
  
  Py_Finalize( );
  return 0;
}
```

I have tried a similar example using std::fstream to modify the file, and this doesn't seem to have the same problems.

hotgloupi:
Maybe this is a typo, but there is no "b" mode for [fopen()](https://msdn.microsoft.com/en-us/library/z5hh6ee9.aspx).

hotgloupi:
It might be that std::fstream uses [_fsopen()](https://msdn.microsoft.com/en-us/library/8f30b0db.aspx) internally. In any case I would expect that exec_file() releases the file handle.

asies:
The extra b tells the file to be opened in binary mode (https://msdn.microsoft.com/en-us/library/z5hh6ee9.aspx).
In any case, I think I also tested the example with the following changes and got the same behaviour:
* opening the file with just "w" instead of "wb"
* writing with sizeof(char) instead of sizeof(byte)

So hopefully that detail isn't critical to the example.


[...]

PR #53, opened 2016-01-08
Close the file handle used in exec_file() (for #52)

hotgloupi:
For python >= 3, a plain `FILE*` is used, whereas for python2, an object handle takes care of the file object lifetime.
This commit manually closes the file handle after the `PyRun_File()` call.

### DIFF
--- a/src/exec.cpp
+++ b/src/exec.cpp
@@ -101,6 +101,9 @@ object BOOST_PYTHON_DECL exec_file(str filename, object global, object local)
                 f,
                 Py_file_input,
 		global.ptr(), local.ptr());
+#if PY_VERSION_HEX >= 0x03000000
+  fclose(fs);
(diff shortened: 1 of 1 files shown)


paralin:
I still have this problem, can this be merged soon?

paralin:
Actually for some reason this causes an access violation:

` 0xC0000005: Access violation writing location 0x00000014.`

So.. maybe doesn't work quite right :)

hotgloupi:
Can you reproduce easily ? Do you have a backtrace ?

paralin:
@hotgloupi I undid the patch and don't have a backtrace. Under python 3.4 however it seems the file descriptor was cleaned up already, resulting in multiple fdclose calls on the descriptor. I didn't have too much time to look into it.

PR #55, opened 2016-02-10
Weaken the precondition of registry insertion in Boost.Python.

camio:
Instead of prohibiting multiple registrations of the same type, we only
prohibit (but don't check) that multiple registrations of the same type
don't conflict.

This enables the same 'class_' call to be used in different scopes as long
as the (semantically) same converter is used.

### DIFF
--- a/src/converter/registry.cpp
+++ b/src/converter/registry.cpp
@@ -209,20 +209,6 @@ namespace registry
 #  endif 
       entry* slot = get(source_t);
       
-      assert(slot->m_to_python == 0); // we have a problem otherwise
-      if (slot->m_to_python != 0)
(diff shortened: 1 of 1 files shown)


stefanseefeld:
I believe I just answered this request on the mailing list: https://mail.python.org/pipermail/cplusplus-sig/2016-April/017395.html

Having multiple registrations for C++ <-> Python type conversions isn't possible right now. There has been discussion on tying this to individual modules, rather than having this an application-global scope, but implementing this is a lot of work.
However, in this case I don't even see the need to support multiple registrations, as the problem can be solved more elegantly with a different approach anyhow.

PR #64, opened 2016-04-28
Assert refcount before decrement

WKarel:
Due to programming errors, the reference count may already be zero before trying to decrement it. If nevertheless, the reference count is decremented, the PyObject will be left with a large reference count, pointing to already destroyed memory, possibly resulting in a hard-to-track segmentation fault. This fix introduces assertions that the reference count is above zero before decrementing it, which provide useful indications of this programming error in debug-builds (even when using a release-build of Python).

This [Gist](https://gist.github.com/WKarel/ff9eff5be792e0b4af2f4d12098a1bcb) shows how to produce a situation in which the assertions are helpful.

### DIFF
--- a/include/boost/python/object_core.hpp
+++ b/include/boost/python/object_core.hpp
@@ -422,6 +422,7 @@ inline api::object_base& api::object_base::operator=(api::object_base const& rhs
 
 inline api::object_base::~object_base()
 {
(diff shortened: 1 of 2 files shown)


stefanseefeld:
Thanks for the patch !

PR #76, opened 2016-07-15
Null from override call

mathstuf:
---
Ok, so here's a test, but it doesn't actually test the code I changed. The problem occurs when the `PyFooOverride` returns `NULL`. I can't seem to get it to actually do so (I think there's too much crap in between which does other things). The original problem was, I think, something gone terribly wrong inside the Python interpreter, but the code path *is* possible (never would have found this if not).

### DIFF
--- a/include/boost/python/override.hpp
+++ b/include/boost/python/override.hpp
@@ -96,12 +96,16 @@ class override : public object
     detail::method_result
     operator()() const
     {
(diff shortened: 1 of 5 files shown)


mathstuf:
I can't reproduce the problem anymore; I guess it has been fixed in a different way since then? Closing.

PR #77, opened 2016-07-21
Multiple calling conventions with MSVC14/amd64

cowo78:
Avoid multiple template instances of boost::python::details::get_signature when multiple calling conventions are enabled


stefanseefeld:
This looks good. Thanks for working on it !

cowo78:
The added ifdef should be fairly general, but I have no idea about why the 32bit target compiler did not complain about multiple definitions.

Issue #70, opened 2016-05-31
/test/exec.cpp - it is wrong to call PyImport_AppendInittab() after Py_Initialize()


vmurashev:
The test in /test/exec.cpp looks like 
1) Py_Initialize() is called in main()
2) far lately in exec_test() PyImport_AppendInittab() is called to inject module 'embedded_hello'
It is wrong and doesn't work in python3.5
---
According to to discussion in python.devel mailing list ( see http://permalink.gmane.org/gmane.comp.python.devel/157983 ) it is wrong to call PyImport_AppendInittab() after Py_Initialize()

Any embedded module injected by PyImport_AppendInittab() and after  Py_Initialize() can't be imported in python3.5. Also there is a comment in python sources confirming this fact:
https://github.com/python/cpython/blob/f680b517e2701e9a3859afb62628a46eccdce17c/Python/import.c#L2153
---
Comment essence:
---
API for embedding applications that want to add their own entries
to the table of built-in modules.  This should normally be called
*before* Py_Initialize().  When the table resize fails, -1 is
returned and the existing table is unchanged.

vmurashev:
patch suggested - see pull request 88

stefanseefeld:
Merged, thanks.

Issue #87, opened 2016-09-24
tests for python3 - what about to remove useless dependence on 'past.builtins' ?


vmurashev:
Rationale:
'past.builtins' is not a part of stock python modules, as result it's difficult to run boost-python tests in limited environments (like Android) at the same time it is the one line only to fix it -
Code like
if (sys.version_info.major >= 3):
    from past.builtins import long

Can be easily replaced with:
if (sys.version_info.major >= 3):
    long = int

vmurashev:
patch suggested - see pull request 88

stefanseefeld:
Much simpler indeed. Merged, thanks !

PR #88, opened 2016-09-24
Fix issues #70, #87

vmurashev:


stefanseefeld:
Thanks for the patch !

PR #91, opened 2016-10-21
exec/eval(): add overloads for char const*

jzwinck:
Many times the caller may have a string created in C++,
so there is no need to wrap it in a Python object when
the only thing done with the object is extract<char*>.

### DIFF
--- a/doc/tutorial.qbk
+++ b/doc/tutorial.qbk
@@ -1417,6 +1417,8 @@ eval evaluates the given expression and returns the resulting value.
 exec executes the given code (typically a set of statements) returning the result,
 and exec_file executes the code contained in the given file.
 
+There are also overloads taking `char const*` instead of str as the first argument.
+
 The [^globals] and [^locals] parameters are Python dictionaries
 containing the globals and locals of the context in which to run the code.
 For most intents and purposes you can use the namespace dictionary of the
--- a/include/boost/python/exec.hpp
+++ b/include/boost/python/exec.hpp
@@ -20,12 +20,20 @@ object
 BOOST_PYTHON_DECL
 eval(str string, object global = object(), object local = object());
 
+object 
+BOOST_PYTHON_DECL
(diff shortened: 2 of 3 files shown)


jzwinck:
@stefanseefeld Perhaps I could ask you to review this?

stefanseefeld:
What problem are you trying to solve with this patch ?

jzwinck:
@stefanseefeld This patch solves the problem that if the string you want to exec or eval is a C string literal (which is a common case), you previously needed to construct a temporary boost::python::str from your C string just to pass it to Python's C API.  Constructing such temporary requires a call to `PyObject_MALLOC()` (via `PyString_FromString()` in Python 2).

In other words, this patch avoids unnecessary memory allocation, copying, and scanning of strings which may be very long.

stefanseefeld:
Just for avoidance of doubt, the following code works fine:
```
boost::python::eval("some python code");
```
i.e., the wrapping of a `char const *` in a python string is implicit.
I understand your concern with overhead due to unnecessary copying of (long) strings, but am not sure to what degree this is a problem in practice. Are you dealing with a case where this incurs measurable overhead, or is this a theoretical concern only ?

jzwinck:
@stefanseefeld Yes I am aware of the implicit conversion.  But as I'm sure you know, C++ and Boost users are quite accustomed to only paying for what they need, and not incurring unnecessary runtime overhead.  This patch is very straightforward, and does not make the library harder to use or understand, so I do not feel the need to justify it with performance benchmarks.  Avoiding unnecessary allocation, copying, and scanning of strings whose lengths are not knowable to us should be reason enough to consider this simple patch.

jzwinck:
@stefanseefeld Would you please consider merging this?  If not I intend to close it.

stefanseefeld:
Yes, done. Sorry it too me so long ! 
(In fact, I think I'm going to deprecate the original functions as there is little use in passing python `str` objects when starting from the C++ side.)

PR #92, opened 2016-11-17
Updated python library reference url for pickle

al3xst:
Due to changes on the python website, the url to the pickle library reference page was invalid.

Edit: Retrospectively all urls in the older boost docs don't work correctly.


stefanseefeld:
Thanks !

PR #93, opened 2016-12-13
Fix conversion of PyUnicodeObject to wstring

tadeu:
On Windows, with Python >= 3.3, `PyObject_Length` cannot be used to get
the size of the `wchar_t` string, because it will count the number of
*code points*, but some characters not on the BMP will use two UTF-16
*code units* (surrogate pairs).
This is not a problem on Unix, since `wchar_t` is 32-bit.

This also fixes a problem where `test_builtin_converters.py` was not
being run, since the module docstring was not the first statement.

### DIFF
--- a/src/converter/builtin_converters.cpp
+++ b/src/converter/builtin_converters.cpp
@@ -430,6 +430,22 @@ namespace
       // Remember that this will be used to construct the result object 
       static std::wstring extract(PyObject* intermediate)
       {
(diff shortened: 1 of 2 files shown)


tadeu:
@stefanseefeld Of course, I've just made the changes, thanks!

stefanseefeld:
Many thanks !

Issue #98, opened 2017-01-23
Building on Windows does not create shared Boost.Numpy


dolfim:
At conda-forge/boost-feedstock#32 we are trying to finalize the Boost Python package on Conda.
The build currently fails while copying ```boost_numpy-vc90-mt-1_63.lib```, because the file does not exist.

The full log is at: boost_numpy-vc90-mt-1_63.lib
The build command is:
```batch
.\b2 install ^
    --build-dir=buildboost ^
    --prefix=%LIBRARY_PREFIX% ^
    toolset=msvc-%VS_MAJOR%.0 ^
    address-model=%ARCH% ^
    variant=release ^
    threading=multi ^
    link=static,shared ^
    --with-python ^
    -j%CPU_COUNT%
```

Is there something wrong in the build process, or did we find an issue in the Boost building process?

jakirkham:
I think the link to the log got lost. Linked below.

Note: This seems to happen with all Boost.Python builds on Windows, but not on Unix.

ref: https://ci.appveyor.com/project/conda-forge/boost-feedstock/build/1.0.97

jakirkham:
Any help on this would be greatly appreciated. We're kind of at a loss ATM.

jhunold:
It is the other way round. Building the shared libraries fails as the missing `lib`-prefix indicates the shared library. This is no wonder as the necessary dllexport/-import magic is completely missing from Boost.Numpy. The msvc thus refuses to create the empty `boost_numpy-vc90-mt-1_63.lib' stub which is needed by the `install`routine.

jakirkham:
Yeah, sorry, we just realized we had it backwards earlier.

So what sort of change would we be looking at?

XWarin:
Hi !

I stumbled on that issue yesterday  trying to get rid of the old (independent) boost numpy.
Is this there something scheduled to correct this dynamic link problem ?
Best regards

SPKorhonen:
I also stumbled onto this problem when trying to update boost to 1.63 to get rid of separate boost::numpy dependency. 

The problem is that, at least on windows, numpy generates empty shared libraries (the dll is generated but exports no symbols). This is due to the windows dll default settings. Everything that is not explicitly exported is private. As @jhunold stated the numpy is missing dllexport/-import magic words. I think that liberal sprinkling of BOOST_PYTHON_DECL before all symbols that are to be exported would help.

SPKorhonen:
[...]

PR #110, opened 2017-02-15
Fix for missing export symbols, issue #98

SPKorhonen:
Fix for missing export symbols in shared library of boost::python::numpy as reported and discussed in issue #98. Added config.hpp modified from python/detail/config.hpp for also numpy. Also added export declaration to functions/classes,

### DIFF
--- a/include/boost/python/numpy/config.hpp
+++ b/include/boost/python/numpy/config.hpp
@@ -12,6 +12,8 @@
 #ifndef CONFIG_NUMPY20170215_H_
 # define CONFIG_NUMPY20170215_H_
 
+# include <boost/config.hpp>
+
 /*****************************************************************************
  *
  *  Set up dll import/export options:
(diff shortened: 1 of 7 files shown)


include/boost/python/numpy/config.hpp
```diff
@@ -0,0 +1,79 @@
...
+//  producing this work.
+
+//  Revision History:
+//  15 Feb 17  Initial version
+
+#ifndef CONFIG_NUMPY20170215_H_
+# define CONFIG_NUMPY20170215_H_
+
+# include <boost/config.hpp>
+# include <boost/detail/workaround.hpp>
```
stefanseefeld: What is this header needed for ?

SPKorhonen: It is not used anymore. I just tested it on my build rig. Submitted new version.

SPKorhonen:
You're welcome! Good to have this fixed. 

PS. Sorry for the extra steps on the way!

Issue #113, opened 2017-02-17
Function checked_downcast_impl declared but not defined


SPKorhonen:
Header pyobject_type.hpp defines function checked_downcast_impl but it is not defined anywhere as far as I can see. It is used by pyobject_type::checked_downcast, which is further referenced by boost::numpy in header numpy_object_mgr_traits.hpp.

Function checked_downcast_impl should be defined or removed.

stefanseefeld:
This was done in https://github.com/boostorg/python/commit/edd890bd2bdb89ac0d9fbcf5b6ba1dcc0bb66717.

PR #114, opened 2017-02-17
Added definition for function checked_downcast_impl

SPKorhonen:
Header pyobject_type.hpp defines function checked_downcast_impl but it is not defined anywhere as far as I can see. It is used by pyobject_type::checked_downcast, which is further referenced by boost::numpy in header numpy_object_mgr_traits.hpp.

Also see issue #113. Causes problems under Windows when using shared libraries


stefanseefeld:
This is obsoleted by https://github.com/boostorg/python/commit/edd890bd2bdb89ac0d9fbcf5b6ba1dcc0bb66717.

PR #115, opened 2017-02-17
Further fix for dll export under windows

SPKorhonen:
There was still some issues using boost::python::numpy under windows as a shared library. At least under MSVC using builtin dtypes is fragile and creates semi-random segfaults. 

This pull request fixes the remaining export issues and fixes the segfault in dtype

### DIFF
--- a/include/boost/python/numpy/dtype.hpp
+++ b/include/boost/python/numpy/dtype.hpp
@@ -56,7 +56,7 @@ public:
    *  This is more permissive than equality tests.  For instance, if long and int are the same
    *  size, the dtypes corresponding to each will be equivalent, but not equal.
    */
-  friend bool equivalent(dtype const & a, dtype const & b);
+  friend BOOST_NUMPY_DECL bool equivalent(dtype const & a, dtype const & b);
 
   /**
    *  @brief Register from-Python converters for NumPy's built-in array scalar types.
@@ -70,7 +70,7 @@ public:
 
 };
 
-bool equivalent(dtype const & a, dtype const & b);
+BOOST_NUMPY_DECL bool equivalent(dtype const & a, dtype const & b);
 
 namespace detail
 {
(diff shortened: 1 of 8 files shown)


include/boost/python/numpy/dtype.hpp
```diff
@@ -70,37 +70,37 @@ class BOOST_NUMPY_DECL dtype : public object {
 
 };
 
-bool equivalent(dtype const & a, dtype const & b);
+BOOST_NUMPY_DECL bool equivalent(dtype const & a, dtype const & b);
 
 namespace detail
 {
```
SPKorhonen: Sorry for the delay. I seem to have missed the notification about your comment.

Anyways, the most pertinent change is the sprinkling of several "BOOST_NUMPY_DECL" statements to force the generation of still missing symbols and objects.

The other changes were necessitated by compilation problems. Under MSVC there were several chicken-or-the-egg -problems with the caused need to apply BOOST_NUMPY_DECL which was problematic due to the forward declarations and such. Also the whole dtype engine seemed to be rather fragile and caused semi-random segfaults depending on the builds. 

After a moderate diagnosis effort I noticed that if you re-arrange the code and especially the declarations in certain logical order the compiler is happy and you can simplify the code and get rid of the segfaults. So yes, apart from BOOST_NUMPY_DECL, the patch is mostly re-arranging the code, but it is done to keep compiler happy.

I've been running this code under several applications and extreme load. As far as I can tell the DLLs produced appear to be rock solid.

stefanseefeld:
Could you separate this PR into one containing thi missing declspec annotations, and another with the remainder of the original patch ? Thanks.

SPKorhonen:
I can. 

For the declspec annotations patch, should I leave the BOOST_NUMPY_DECL out where I can't add them without breaking the build on MSVC thus leaving the dtype in inconsistent state or should I add them regardless of the effect on MSVC build?

stefanseefeld:
I believe all the issues your PR attempted to fix are now addressed. If not, please update your PR and resubmit.

PR #126, opened 2017-05-08
Fixing compiling error 'error: ‘NPY_FLOAT16’ was not declared in this scope'

hamparawa:
This issue is already fixed in ndarray/Boost.NumPy
https://github.com/ndarray/Boost.NumPy/issues/58


stefanseefeld:
Thanks !

PR #130, opened 2017-06-02
Reencoded a few headers that used Windows-1252 with UTF-8.

Teemperor:
Nearly every header in the boost codebase is UTF-8, but here there
are a few headers which are using Windows-1252, which makes it impossible
for some tools to parse those files. This patch just reencodes them
with UTF-8 like the rest of the codebase. I checked that the name of the
author is still correct after this change.

No functional change intended.


stefanseefeld:
This looks good; thanks !
I'm waiting for travis-ci results, though (which seems particularly slow today).

Issue #127, opened 2017-05-10
Error when compiling client code "could not find boost_python-vc140-mt-1_64.lib"


youcefl:
When compiling Boost libraries I used to get two lib/dll files for Boost.Python e.g.:
boost_python-vc120-mt-1_60.lib and boost_python3-vc120-mt-1_60.lib
(even though I only targeted Python 3.4), now, building version 1.64, I only get
boost_python3-vc140-mt-1_64.lib so the automatic linking does not work anymore
because the file <boost/python/detail/config.hpp> only has
#define BOOST_LIB_NAME boost_python
Is it possible to restore the old behavior ? I could create a symbolic link
boost_python-vc140-mt-1_64.lib -> boost_python3-vc140-mt-1_64.lib
but I'd prefer not having to do so.

jakirkham:
Also ran into this issue as well.

Was discussing this with a colleague and we agree that the issue is likely this [line]( https://github.com/boostorg/python/blob/boost-1.64.0/include/boost/python/detail/config.hpp#L108 ). It should probably be replaced with code like this.

```c++
#if PY_MAJOR_VERSION < 3
# define BOOST_LIB_NAME boost_python
#else
# define BOOST_LIB_NAME boost_python3
#endif
```

Thoughts on this @stefanseefeld?

jakirkham:
Might also be worth checking a few [other places]( https://github.com/boostorg/python/search?utf8=%E2%9C%93&q=boost_python&type= ) just to be sure.

stefanseefeld:
Thanks for investigating this. I'm neither a Windows developer nor a Windows user, so this issue is somewhat outside my own even horizon.
Yes, the problem is the above line as far as I understand. However, it isn't entirely clear to me what the "right" fix is. Do we really want to hardcode the name of the library based on what Python version is used ? I think it might be better to keep the hardcoded definition but conditionalize it:
```
#ifndef BOOST_LIB_NAME
# define BOOST_LIB_NAME boost_python
#endif
```
and then let the build system provide its macro to override this default.

jakirkham:
Sorry I'm unclear on this. What build system is defining `BOOST_LIB_NAME`? Also why shouldn't this be something Boost.Python provides?

stefanseefeld:
[...]

PR #135, opened 2017-06-10
Fix `BOOST_LIB_NAME` for Python 3

jakirkham:
Fixes https://github.com/boostorg/python/issues/127

This was reusing the Python 2 name on Python 3, which is incorrect since the Python 3 library for Boost.Python now has a `3` in it. Hence this checks against the Python version and defines this correctly.

### DIFF
--- a/include/boost/python/detail/config.hpp
+++ b/include/boost/python/detail/config.hpp
@@ -105,7 +105,11 @@
 // Set the name of our library, this will get undef'ed by auto_link.hpp
 // once it's done with it:
 //
-#define BOOST_LIB_NAME boost_python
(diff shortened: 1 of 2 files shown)


stefanseefeld:
As I just explained in https://github.com/boostorg/python/issues/127, I'm not comfortable with the idea that the library name is hardcoded like this into the code. The build system is a better place for this. For that reason I'd favour a making the hardcoded name a default only (i.e. prefixed witth `#ifndef BOOST_LIB_NAME ...`), and then add the above logic to the Jamfiles and wherever else the library name is being defined.

Issue #106, opened 2017-02-08
Replace use of boost type-traits by std type-traits.


stefanseefeld:
While this is a good idea in general, this also fixes certain issues (e.g. https://github.com/boostorg/python/issues/105).

This may require some code refactoring, to avoid the code being cluttered with #ifdefs.

shreyans800755:
@stefanseefeld @dabrahams 
I would like to work on this issue.

Just to confirm and get along with contributing method, I need to create PR against `develop` branch, right ?

stefanseefeld:
@shreyans800755 , great ! I'm not aware of anyone else working on this right now, so this would be a great contribution. Please be sure to conditionalize these substitutions to only be active whenever a compiler with the appropriate language support is used. We need to maintain backward compatibility.

I suggest a new `boost/python/detail/type_traits.hpp` header for this.

PRs need to be created against the develop branch.

shreyans800755:
@stefanseefeld 
I'm trying to apply the patch like following:
Adding
```
#if __cplusplus < 201103L
          typename boost::remove_cv<T>::type
#else
          typename std::remove_cv<T>::type
#endif
```
and removing
`typename boost::remove_cv<T>::type`

I don't understand why we need a new file `boost/python/detail/type_traits.hpp`. Can you please explain ? Am I missing something ?

stefanseefeld:
I expect that with the above approach you would end up with lots of locations in the code that would use such a conditional logic, which IMO makes the code hard to read.

Thus what I'd propose is something like:
```
namespace boost { namespace python { namespace detail {
#if __cplusplus < 201103L
  using boost::remove_cv;
  ...
#else
  using std::remove_cv;
  ..
}}} // namespace boost::python::detail
```
such that in the rest of Boost.Python's code you could then simply refer to `boost::python::detail::remove_cv` etc., not having to care which original definition this aliases.
(Note, however, that this is just an idea that I would try out. I haven't actually tested this, and thus could be convinced that it isn't workable. I'd put the conditional above into a new header `boost/python/detail/type_traits.hpp`...)

shreyans800755:
That's wonderful idea. I knew, changes were a bit cumbersome. This sounds like an elegant solution. I'll try to move ahead with this.

PR #136, opened 2017-06-24
Use std type_traits instead of boost type_traits

shreyans800755:
Fixes https://github.com/boostorg/python/issues/106

### DIFF
--- a/include/boost/python/detail/type_traits.hpp
+++ b/include/boost/python/detail/type_traits.hpp
@@ -1,0 +1,110 @@
+// Copyright Shreyans Doshi 2017.
+// Distributed under the Boost Software License, Version 1.0. (See
+// accompanying file LICENSE_1_0.txt or copy at
+// http://www.boost.org/LICENSE_1_0.txt)
+
+#ifndef BOOST_PYTHON_DETAIL_TYPE_TRAITS_HPP
+# define BOOST_PYTHON_DETAIL_TYPE_TRAITS_HPP
+
+
+#if __cplusplus < 201103L
+# include <boost/type_traits/transform_traits.hpp>
+# include <boost/type_traits/same_traits.hpp>
+# include <boost/type_traits/cv_traits.hpp>
+# include <boost/type_traits/is_polymorphic.hpp>
+# include <boost/type_traits/composite_traits.hpp>
+# include <boost/type_traits/conversion_traits.hpp>
+# include <boost/type_traits/add_pointer.hpp>
+# include <boost/type_traits/remove_pointer.hpp>
+# include <boost/type_traits/is_void.hpp>
+# include <boost/type_traits/object_traits.hpp>
+# include <boost/type_traits/add_lvalue_reference.hpp>
+# include <boost/type_traits/function_traits.hpp>
+# include <boost/type_traits/is_scalar.hpp>
+# include <boost/type_traits/alignment_traits.hpp>
+# include <boost/mpl/bool.hpp>
+#else
+# include <type_traits>
+#endif
+
+# include <boost/type_traits/is_base_and_derived.hpp>
+# include <boost/type_traits/alignment_traits.hpp>
+# include <boost/type_traits/has_trivial_copy.hpp>
+
+
+namespace boost { namespace python { namespace detail {
+
+#if __cplusplus < 201103L
+    using boost::alignment_of;
+    using boost::add_const;
+    using boost::add_cv;
+    using boost::add_lvalue_reference;
+    using boost::add_pointer;
+
+    using boost::is_array;
+    using boost::is_class;
+    using boost::is_const;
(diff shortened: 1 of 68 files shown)


include/boost/python/detail/type_traits.hpp
```diff
@@ -0,0 +1,105 @@
+// Copyright David Abrahams 2002.
```
stefanseefeld: Shouldn't that be your name, and '2017' ? :-)

include/boost/python/detail/type_traits.hpp
```diff
@@ -0,0 +1,105 @@
...
+// Distributed under the Boost Software License, Version 1.0. (See
+// accompanying file LICENSE_1_0.txt or copy at
+// http://www.boost.org/LICENSE_1_0.txt)
+
+#ifndef BOOST_DETAIL_TYPE_TRAITS_HPP
+# define BOOST_DETAIL_TYPE_TRAITS_HPP
```
stefanseefeld: This should be BOOST_PYTHON_DETAIL_TYPE_TRAITS_HPP, to really match the complete filename.

include/boost/python/detail/type_traits.hpp
```diff
@@ -0,0 +1,105 @@
...
+# include <boost/type_traits/conversion_traits.hpp>
+# include <boost/type_traits/add_pointer.hpp>
+# include <boost/type_traits/remove_pointer.hpp>
+# include <boost/type_traits/is_void.hpp>
+# include <boost/type_traits/object_traits.hpp>
+# include <boost/type_traits/add_lvalue_reference.hpp>
+# include <boost/type_traits/function_traits.hpp>
+# include <boost/type_traits/is_scalar.hpp>
+# include <boost/type_traits/alignment_traits.hpp>
+# include <boost/mpl/bool.hpp>
```
stefanseefeld: Have you tried only conditionally including the above headers, using the same logic as below (i.e., if C++11 isn't supported) ? This would greatly reduce our dependency on Boost...

include/boost/python/detail/value_is_xxx.hpp
```diff
@@ -9,23 +9,21 @@
...
+{                                                                                              \
+    BOOST_PYTHON_IS_XXX_DEF(name,qualified_name,nargs)                                         \
+    BOOST_STATIC_CONSTANT(bool, value = is_##name<                                             \
+                               typename boost::python::detail::remove_cv<                      \
+                                  typename boost::python::detail::remove_reference<X_>::type   \
+                               >::type                                                         \
+                           >::value);                                                          \
+    typedef mpl::bool_<value> type;                                                            \
+                                                                                               \
 };
```
stefanseefeld: It looks like right now (i.e. even before your change) the `value_is_...` template is defined in the global namespace. Can it be moved into boost::python::detail ?

include/boost/python/detail/type_traits.hpp
shreyans800755: Ah yes! I just copied it from some other file blindly.

include/boost/python/detail/type_traits.hpp
```diff
@@ -0,0 +1,110 @@
...
+#else
+# include <type_traits>
+#endif
```
stefanseefeld: Great, I hope with this change it builds on Travis-CI at least.

include/boost/python/detail/value_is_xxx.hpp
```diff
@@ -9,24 +9,25 @@
...
+namespace boost { namespace python { namespace detail {
+
+#  define BOOST_PYTHON_VALUE_IS_XXX_DEF(name, qualified_name, nargs)                           \
+template <class X_>                                                                            \
+struct value_is_##name                                                                         \
+{                                                                                              \
+    BOOST_PYTHON_IS_XXX_DEF(name,qualified_name,nargs)                                         \
+    BOOST_STATIC_CONSTANT(bool, value = is_##name<                                             \
+                               typename boost::python::detail::remove_cv<                      \
+                                  typename boost::python::detail::remove_reference<X_>::type   \
```
stefanseefeld: Last minor nit-pick: now that this entire template is within the boost::python::detail namespace, it appears we don't need the qualification of `remove_cv` and `remove_reference` in the last two lines any longer.

shreyans800755: I'll update it. Btw any idea, why appveyor is failing ?

stefanseefeld: Great, thanks. Don't worry about appveyor, that's simply a problem with the testing logic (SCons). I'm in the process of substituting that by something else...

shreyans800755: Cool.

PR #139, opened 2017-07-12
Fix wrong and unused macro name while use BOOST_PP

yocox:
- `BOOST_PYTHON_AS_OBJECT` in numeric.cpp has never been used
- `BOOST_PYTHON_AS_OBJECT` in numeric.hpp should be `BOOST_PYTHON_ENUM_AS_OBJECT`


stefanseefeld:
In fact, I believe now that we have proper NumPy support in Boost.Python, we should deprecate the Numeric API. Thoughts ?

Issue #140, opened 2017-07-17
''::hypot' has not been declared' and 'undefined reference to `__imp_Py_InitModule4'' errors for mingw-w64 builds on Windows


sav-ix:
Hello everyone,

For Boost.Python builds using mingw-w64 on Windows got error:
```
.\b2 --build-type=complete --with-python toolset=gcc-6.3.0 variant=release link=shared threading=multi runtime-link=shared address-model=64 -j1 -d2 -q

<snip>

gcc.compile.c++ bin.v2\libs\python\build\gcc-mingw-6.3.0\release\threading-multi\numeric.o

    "g++"  -ftemplate-depth-128 -O3 -finline-functions -Wno-inline -Wall -mthreads -m64  -DBOOST_ALL_NO_LIB=1 -DBOOST_PYTHON_SOURCE -DNDEBUG  -I"." -I"C:\ProgramData\Python64\Include" -c -o "bin.v2\libs\python\build\gcc-mingw-6.3.0\release\threading-multi\numeric.o" "libs\python\src\numeric.cpp"

In file included from C:/ProgramData/Mingw64/lib/gcc/x86_64-w64-mingw32/6.3.0/include/c++/math.h:36:0,
                 from C:\ProgramData\Python64\Include/pyport.h:325,
                 from C:\ProgramData\Python64\Include/Python.h:58,
                 from ./boost/python/detail/wrap_python.hpp:151,
                 from ./boost/python/detail/prefix.hpp:13,
                 from ./boost/python/numeric.hpp:8,
                 from libs\python\src\numeric.cpp:6:
C:/ProgramData/Mingw64/lib/gcc/x86_64-w64-mingw32/6.3.0/include/c++/cmath:1157:11: error: '::hypot' has not been declared
   using ::hypot;
           ^~~~~

<snip>

...failed gcc.compile.c++ bin.v2\libs\python\build\gcc-mingw-6.3.0\release\threading-multi\numeric.o...
...failed updating 1 target...
...updated 8 targets...
```

Reproduced for builds using mingw-w64 with `<Debug|Release> <Shared|Static>` configurations. Not reproduced for builds using ICC or MSVC.

Could be fixed by adding `cxxflags="-D_hypot=hypot"` key to Boost build commands (see [appropriate topic](http://stackoverflow.com/questions/10660524/error-building-boost-1-49-0-with-gcc-4-7-0)).

If blocker above is fixed, next error appear:
```
[...]

PR #141, opened 2017-07-20
Fix Python-related redefinitions for builds using mingw-w64

sav-ix:
Fixes https://github.com/boostorg/python/issues/140.

### DIFF
--- a/build/Jamfile
+++ b/build/Jamfile
@@ -36,6 +36,8 @@ py3-version = [ py-version 3 ] ;
 
 project boost/python
   : source-location ../src
+  : requirements <toolset>gcc <target-os>windows:<cxxflags>-D_hypot=hypot
+  : requirements <toolset>gcc <target-os>windows <link>shared:<cxxflags>-DMS_WIN64
   ;
 
 rule cond ( test ? : yes * : no * ) { if $(test) { return $(yes) ; } else { return $(no) ; } }


stefanseefeld:
By "broke MSVC builds", are you referring to the Appveyor build failure ? That's an unrelated issue with the SCons build logic. Boost.Build is not involved in the CI builds. If not, can you show an error log ? (I'm only a casual Windows user, so I may not be able to help.)

sav-ix:
Travis failures forced me to test this patch for MSVC builds. Which I missed to do before publishing this PR.
For builds using MSVC with Debug+Static and Release+Shared configurations got the same errors:
```
error: "gcc" is not a known value of feature <toolset>
error: legal values: "msvc"
```
(see [logs](https://github.com/boostorg/python/files/1162258/boost_msvc_build_logs.zip) in attachment).

build/Jamfile
```diff
@@ -36,6 +36,8 @@ py3-version = [ py-version 3 ] ;
 
 project boost/python
   : source-location ../src
+  : requirements <toolset>gcc <target-os>windows:<cxxflags>-D_hypot=hypot
+  : requirements <toolset>gcc <target-os>windows <link>shared:<cxxflags>-DMS_WIN64
```
stefanseefeld: I don't think this is doing what you expect. As far as I understand Boost.Build syntax, the last line above defines as requireents:
* toolset=gcc
* target-os=windows
* cxxflags=-DMS_WIN64, whenever link=shared

It looks like what you intend to do is to define -DMS_WIN64 whenever all three features (toolset, target-os, and link) have the given values, yes ? 
Please ask on the boost or boost.build mailing list what the right syntax for this is.
(I'm also not sure how this would fix the `"gcc is not a known value..." error. That's another thing to ask on boost.build ML).
Sorry I can't help more with this. Boost.Build is complicated and badly documented, which is one reason why I'm trying to move away from it...

sav-ix:
I was trying to add rules to Boost.Python build system:
IF build using `mingw-w64` THEN add `-D_hypot=hypot` to compiler keys,
IF build using `mingw-w64` AND layout `shared` THEN add `-DMS_WIN64` to compiler keys,
based on other Jamfiles as samples. But Boost.Build was more complicated then expected.

> Please ask on the boost or boost.build mailing list what the right syntax for this is.

Or we can invite Boost.Build Developers to join this discussion.
@raffienficiaud, could you give a tip, how to solve this issue correctly?

eromoe:
@sav-ix  Would you mind to explain how to add `-D_hypot=hypot`   .  I got similar error while install https://github.com/pytries/DAWG

PR #144, opened 2017-08-04
replace CRLF to LF

joelin:
i found this error when i checkout branch svn-branches/system at my linux computer. 
`modified:   build/VisualStudio/boost_python.dsp` 
the reason is CRLF. 
I synchronizing the all source repo(include all branch and tags ) to intranet and then found this error, it breaks the  job of  synchronize. so replace it to LF.


stefanseefeld:
That's an unmaintained branch, and I'm tempted to remove it (as well as many other similar ones). Can't you limit the branches to sync to, say, "master" and "develop" ?

joelin:
@stefanseefeld thx! I synchronized all branches because i don't know which one is need by another module, eg. ceph. i can limit the branches via customized sync job. it's going to be complicated, that's not what I want. so the best way is delete it or modify it corrected.

PR #145, opened 2017-08-15
Fix warning: returning the proper type for object sizes

raffienficiaud:
This fixes a small casting warning, and I believe is more correct. I however do not know if this is compatible with python < 2.7


raffienficiaud:
Appveyor seems to be broken

PR #146, opened 2017-08-21
Fix autoptr/unique_ptr ifdef for VS2017.

garyfurnish:
This fixes autoptr detection on VS2017.

### DIFF
--- a/include/boost/python/to_python_indirect.hpp
+++ b/include/boost/python/to_python_indirect.hpp
@@ -83,7 +83,7 @@ namespace detail
           // copy constructor.
 # if defined(__ICL) && __ICL < 600 
           typedef boost::shared_ptr<T> smart_pointer;
-# elif __cplusplus < 201103L
+# elif defined(BOOST_NO_CXX11_SMART_PTR)
           typedef std::auto_ptr<T> smart_pointer;
 # else
           typedef std::unique_ptr<T> smart_pointer;
--- a/config/cxx.py
+++ b/config/cxx.py
@@ -15,7 +15,7 @@ def add_options(vars):
 
 def check(context):
 
-    source = r"""#if __cplusplus < 201103L
+    source = r"""#if __cplusplus < 201103L && (!defined(_MSC_VER) || _MSC_VER < 1800) 
 #error no C++11
 #endif"""
 
--- a/include/boost/python/converter/registered.hpp
+++ b/include/boost/python/converter/registered.hpp
@@ -76,7 +76,7 @@ namespace detail
       registry::lookup_shared_ptr(type_id<shared_ptr<T> >());
   }
 
-#if __cplusplus >= 201103L
+#if !defined(BOOST_NO_CXX11_SMART_PTR)
   template <class T>
   inline void
   register_shared_ptr0(std::shared_ptr<T>*)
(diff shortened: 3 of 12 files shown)


stefanseefeld:
Can you elaborate a bit on a) the rationale for this patch and b) the symptoms you are observing ?
I'm not entirely convinced of the patch, at least not without further (inline) comments. The macro discriminates the availability of smart pointers but the conditional code involves `std::move`. There is no apparent relationship between the two (other than that both are C++11 features)...

garyfurnish:
The rationale is that auto_ptr is permanently removed from MS stl as of VS2017 update 3 per c++ standard, so any auto_ptr use is an unrescuable problem.  You can't copy unique pointers so initialization must be done via move.  I suspect this subtly breaks other things.

stefanseefeld:
And VS2017 doesn't set `__cplusplus` properly ?

garyfurnish:
No :(

garyfurnish:
Interestingly enough in spite of the failing integration tests on VS this does compile successfully on my machine with VS 2017 Update 3, so maybe its something that needs a work around that was a silently fixed compiler bug.

garyfurnish:
Is it possible to get more versions of VS on appveyor?

MarcelRaad:
@garyfurnish : the problem only occurs when compiling in C++17 mode via `/std:c++latest` or `/std:c++17`. If you're talking about the official Boost regression test runners, half of the regression test runners are using `/std:c++latest`, and these are failing.

I believe support for `std::move` is checked via `BOOST_NO_CXX11_RVALUE_REFERENCES` most of the time, so maybe both could be used? Even though I don't know what C++11 smart pointers without rvalue references would mean, maybe only `std::shared_ptr` and `std::weak_ptr`...

garyfurnish:
@MarcelRaad Well, there are really a few issues here. 
1) the scons that runs the tests on appveyor doesn't detect c++11 correctly on windows
2) if you fix c++latest to use autoptr, there is an awful linker error.  Somewhere between the version of vc++ on appveyor and 2017 update 3, the error mysteriously stopped being an error, so I'm assuming it was some sort of type system bug.
3) Having fixed both 1 and 2 and running on 2017 update 3, some of the tests on windows are failing because there are other places where c++11 support detection on windows aren't detected correctly.  Will fix this after the eclipse.
4) But fixing the regression errors on appveyor will probably require some sort of dicscrimination on MSVC version.  It would be nice if I knew the exact version where it started working, but I don't have an array of VS machines setup.  I'm thinking defining a BOOST_PYTHON_MSVC_WORKAROUND that when true defaults things to autoptr so as to not break existing code.  

But I think I know how to fix all of the current bugs on std:c++ latest.  I only have 3 tests not passing right now, and they seem straightforward detection errors.  Its just going to take some creativity to figure out how to deal with that VC bug.

garyfurnish:
This appears to work correctly and pass tests on MSVC 2017 update 3 on a real machine with scons beta 3.  I believe the appveyor errors are spurious and were there pre-modifications, so this should be almost good to go.
There are errors on import.  This be a preexisting bug, so I think this is good to go!
For whatever reason travis is hung up on running this, so I did test it on a linux box and it works fine just in case the somewhat obvious changes didn't work right.
Does this need anything else to pass inspection?

include/boost/python/to_python_indirect.hpp
```diff
@@ -83,7 +83,7 @@ namespace detail
           // copy constructor.
 # if defined(__ICL) && __ICL < 600 
           typedef boost::shared_ptr<T> smart_pointer;
-# elif __cplusplus < 201103L
+# elif BOOST_NO_CXX11_SMART_PTR
```
glenfe: #elif defined(BOOST_NO_CXX11_SMART_PTR)

garyfurnish:
Bump.

Issue #131, opened 2017-06-06
initModule symbol not visible in 1.64


rosswhitfield:
I'm compiling a project that sets the default visibility to `hidden` using `CMAKE_CXX_VISIBILITY_PRESET=hidden`. This worked in 1.63 as the `initModule` is set back to `default` [here](https://github.com/boostorg/python/blob/boost-1.63.0/include/boost/python/module_init.hpp#L75-L79), unfortunately after #1 `BOOST_PYTHON_USE_GCC_SYMBOL_VISIBILITY` is no longer defined so the visibility is not getting set back to `default`.

So I'm getting errors like `ImportError: dynamic module does not define init function (init_kernel)` and the equivalent for python 3 `ImportError: dynamic module does not define module export function (PyInit__kernel)`

lukaszforynski:
This seems to be the offending change.. 


```
include/boost/python/object/instance.hpp
@@ -11,7 +11,7 @@
 
 namespace boost { namespace python
 {
-  struct BOOST_PYTHON_DECL_FORWARD instance_holder;
+  struct instance_holder;
 }} // namespace boost::python
 
 namespace boost { namespace python { namespace objects { 
```

Changing it back (*to what was originally intended in #1):

```
include/boost/python/object/instance.hpp
@@ -11,7 +11,7 @@
 
 namespace boost { namespace python
 {
-  struct instance_holder;
+  struct BOOST_PYTHON_DECL instance_holder;

 }} // namespace boost::python
 
 namespace boost { namespace python { namespace objects { 
```

Seems to fix the issue.

PR #151, opened 2017-08-22
Fix for #131 initModule symbol not visible in 1.64

formiaczek:
If a global visibility is set to 'hidden', importing modules that define classes without either 'noimport' or 'init' causes 
errors like:
`ImportError: dynamic module does not define init function`

This seems to be due to some, probably unintended changes from #1.


micbou:
Contrarily to PR https://github.com/boostorg/python/pull/159, this doesn't fix the issue when adding the `-fvisibility=hidden` flag to our project. See PR https://github.com/Valloric/ycmd/pull/848.

PR #155, opened 2017-09-12
Add RAII wrapper for GIL release/acquire in gil.hpp

jzwinck:
Taken from my old Trac ticket: https://svn.boost.org/trac10/ticket/7836 which said:

This ticket follows a discussion I started on the Python C++ SIG mailing list (which in turn followed a brief topic on the Boost developers list).

I submit for inclusion in Boost.Python a pair of classes for releasing and acquiring the Python Global Interpreter Lock (GIL) using RAII semantics.

This code (with different naming conventions but the same logic) has been used successfully by me "in production" and the releaser class is inspired by several examples online that others have until now copy-pasted when needed.


jzwinck:
Closing due to inactivity.

Issue #157, opened 2017-09-19
cpp_ndarray.shape(-1) should behave like python_ndarray.shape[-1] and boost::python::tuple::operator[](-1)


mborgerding:
Currently `ndarray::shape(int n)` returns `get_shape()[n]`, which indexes a raw memory buffer (C-style array). This results in undefined behavior under the python idiom of negative indices.

I propose `shape` ( and similarly `strides`) performs bounds checking and indexing from the back for negatives.

stefanseefeld:
That sounds reasonable. Care to submit a patch ? ;-)

mborgerding:
There is a failed CI build, but I cannot see any resemblence to my changes.

`unresolved external symbol  ... boost::get_pointer<struct test_class<0> ...`

stefanseefeld:
Right, I know. (The problem is that the build system I use is broken, too, as the whole test run fails rather than just flagging the failed test. I'm working on a new build system to supersede the SCons-based one.)
Don't worry about that. I'm just waiting for the Travis build to finish. Your patch looks good, and I'm looking forward to merging it once it's confirmed that the new tests pass and no regressions are introduced. Thanks !

PR #158, opened 2017-09-20
ndarray.shape,strides act more like their python counterparts see #157

mborgerding:
addresses #157
Unit tests added for to verify correct behavior for shape(k),strides(k) with negative indexing.  Verify IndexError raised when k out of bounds.

### DIFF
--- a/include/boost/python/numpy/ndarray.hpp
+++ b/include/boost/python/numpy/ndarray.hpp
@@ -86,11 +86,11 @@ public:
   /// @brief Copy the scalar (deep for all non-object fields).
   ndarray copy() const;
 
-  /// @brief Return the size of the nth dimension.
-  Py_intptr_t shape(int n) const { return get_shape()[n]; }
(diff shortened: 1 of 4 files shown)


mborgerding:
@stefanseefeld , I agree the changes shouldn't be lumped together.  That was not my intent, but github did it on its own (I may've missed something) .  If you feel they should be separated, I can try again, maybe rebasing both changes from the master branch would force github to make them separate PRs.

stefanseefeld:
Yeah, simply move the second one into a separate branch from which you can submit another PR. Sorry it's all a bit formal, but with practice it becomes easier ;-)

mborgerding:
Split this into two PRs. From now on I'll know not to believe that github's "Create New Pull Request" will actually create a _New_ pull request.

stefanseefeld:
FWIW, you could simply have added another commit reverting the previous one (which I could have squash-and-merged as one - so only the first one would have been visible), or you could have force-pushed an update removing the second commit.
But of course, starting over with a new PR works, too. :-)

PR #159, opened 2017-09-20
Fix symbol visibility of init method

### LINKED ISSUE #131: initModule symbol not visible in 1.64

marv:
This was properly an oversight when switching to the BOOST_SYMBOL_*
symbols from Boost.Config in commit 0224f54a (see #1)

Since Boost.Config is already doing the differentiation between
different platforms and compilers we can simplify this bit and just
use BOOST_SYMBOL_EXPORT

This should fix #131

### DIFF
--- a/include/boost/python/module_init.hpp
+++ b/include/boost/python/module_init.hpp
@@ -66,25 +66,9 @@ BOOST_PYTHON_DECL PyObject* init_module(char const* name, void(*)());
 
 #  endif
 
-#  if (defined(_WIN32) || defined(__CYGWIN__)) && !defined(BOOST_PYTHON_STATIC_MODULE)
-
-#   define BOOST_PYTHON_MODULE_INIT(name)                           \
-  void BOOST_PP_CAT(init_module_,name)();                           \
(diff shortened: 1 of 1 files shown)


micbou:
I can confirm this fix the visibility issue. Without this change, we have to downgrade Boost to 1.63.0 to use the `-fvisibility=hidden` flag in our project. See PR https://github.com/Valloric/ycmd/pull/848.

PR #19, opened 2015-05-06
Fix compilation with python 3.0-3.3

Lastique:
Backport commit 3e405b6fd5db5615bbef241763de070118222ca7 to develop branch: Fix exec_file for Python 3 < 3.4. Also fix version check to actually fix compilation with python 3.4.

PR #183, opened 2018-01-17
Backporting fixed boost::python::exec_file to v1.58.0

aambrosano:
Backporting src/exec.cpp from boost 1.59 to 1.58 to fix segmentation fault. It should allow any Ubuntu 16.04 (shipping boost 1.58) user to run boost::python::exec_file without segmentation faults.

The fix is related to #15 and #19.

I guess boost 1.59 has been already tested so there shouldn't be any problem, but in any case [here](https://github.com/boostorg/python/files/1638758/test_script.txt) is a script for testing the change with python 2.7.


stefanseefeld:
I don't quite understand what you are trying to do here. You mention "backporting from 1.59 to 1.58", but you submit a PR against develop, which is in sync with the latest release, i.e. 1.66. If you want to backport anything, shouldn't it be *from* develop, to some other stable branch somewhere else ?

aambrosano:
What I am trying to do is to update only boost 1.58 related commits, but as I see there isn't any branch for old releases, just the "boost_1_xx_0" tags, that's where I got it wrong.

Everything went apparently fine on my side until I got to the pull request phase (I forked and patched starting from boost_1_58_0), but now I would need to merge on top of a "boost_1_58_0" _branch_ in order to update only the source code for the old version. Do you think anything like this is possible?

stefanseefeld:
If I understand correctly you want to patch the 1.58 release (with commits from the `develop` branch). What do you want to do with that patched version ? Will that be the basis for new Ubuntu packages ? Or just your personal build ?
As a start, you may want to fork this repo, then you can add a "1.58" branch to which you can backport (or "cherry-pick") any later commits. As you say, this repo doesn't have stable branches for past releases which would serve similar purposes, it only has tags.
If such branches become useful to more people, we may consider merging them back into this repo. But as of now, I'm hesitant, as this implies additional maintenance work, which I'd prefer to delegate to those who need it (e.g., you).

PR #186, opened 2018-02-09
Fix build with Python 3.7

berolinux:
Python 3.7 changes the return type of _PyUnicode_AsString()
from void* to const char* -- causing the build of boost-python
to fail.

Signed-off-by: Bernhard Rosenkränzer <bero@lindev.ch>


stefanseefeld:
Thanks !

PR #206, opened 2018-05-30
crash when generating sphinx documentation after MSVC upgrade

ulrichard:
Generating sphinx documentation worked fine for our python embedding dll built with boost python.
After the compiler upgrade to VisualStudio 15.7.2  it always leads to a crash.
Boost or python were not changed, only MSVC. Recompiling boost with the new MSVC didn't help.
I could prevent the access violation with this nullptr check.
I didn't research the root cause further from there. There might be a better solution.

### DIFF
--- a/src/object/function_doc_signature.cpp
+++ b/src/object/function_doc_signature.cpp
@@ -116,7 +116,7 @@ namespace boost { namespace python { namespace objects {
 
     const  char * function_doc_signature_generator::py_type_str(const python::detail::signature_element &s)
     {
-        if (s.basename==std::string("void")){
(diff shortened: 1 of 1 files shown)


SPKorhonen:
The underlying cause is reported in issue #207 and fixed by pull request #208.

This pull request masks the problem when accessing doc-strings (via __doc__) attribute but still leaves the wrapped function objects in invalid state which leads to hard to diagnose problems when calling functions. Most often you get spurious errors stating the the call signature is not compatible.

stefanseefeld:
Given that #208 is now merged, this PR should be obsolete. @SPKorhonen can you confirm that, please ?

Issue #207, opened 2018-05-30
Visual Studio 2017 generates invalid code for caller.hpp


SPKorhonen:
Under Python 3.6.5 and Visual Studio 2017 (15.7.2) and optimization level /O2 compilation of boost::python generates invalid code while compiling ".\boost\python\detail\caller.hpp". Namely, initialization of "static const signature_element ret" (line 245->) is erroneously skipped leading to SegFault in python.exe.

stefanseefeld:
I have just merged #208 into the `develop` branch, and intend to include it into the upcoming `1.70` release. Would you mind testing again against `develop`, so I can close this issue ? Thanks !

PR #208, opened 2018-05-30
 Fix static object initialization under Visual Studio 2017

SPKorhonen:
By separating the initialization of "static const signature_element ret" into separate function defined in detail namespace the compiler correctly initializes the static variable.

Fixes issue #207


SPKorhonen:
Added comments and fixed indent from tab to space

Issue #209, opened 2018-06-05
numpy support broken


luca-penasa:
This seems to be a bug affecting several distros. I testes on Manjaro (fresh and updated live) and Ubuntu 18.04 (docker).

Here is the code for reproducing the bug:
https://github.com/luca-penasa/boost-numpy-minimal-bug

it compiles a python module using boost-python and boost-python-numpy
the line of code 
```cpp
np::initialize();
```
causes an error when trying to import the module in python (here from ipython but it is the same from pure python):
```
Python 3.6.5 (default, May 11 2018, 04:00:52) 
Type 'copyright', 'credits' or 'license' for more information
IPython 6.3.1 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import hello
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
RuntimeError: FATAL: module compiled as little endian, but detected different endianness at runtime
---------------------------------------------------------------------------
ImportError                               Traceback (most recent call last)
ImportError: numpy.core.umath failed to import
---------------------------------------------------------------------------
SystemError                               Traceback (most recent call last)
<ipython-input-1-f81fb083bdeb> in <module>()
----> 1 import hello

SystemError: initialization of hello raised unreported exception

In [2]: 

```

Could anybody have a look at this? this is making a lot of my code useless!

stefanseefeld:
As the error indicates, it looks like the boost_numpy3 module was compiled for a different architecture (endianness). Please report this to the packager, i.e. the distro you are using.

luca-penasa:
thank you for the reply @stefanseefeld 
That would also mean that both in the case of ubuntu and arch they are compiling boost with the wrong endianess... It sounds unlikely to me. what do you think? I cannot be sure the problem lies within boost-python-numpy but it seemed to me ther more likely.  ... I've been also trying recompiling it several times with no success. 

On what distro are you running it?
does the sample code compile for you?

I am hitting the same problem both on my workstation and on the laptop...

thanks

luca-penasa:
btw
i am on manjaro (arch) witht the following packages:
python 3.6.5
numpy 1.14.3
boost 1.67.0

I've been tring several versions of boost and numpy (trunk) with no success... any idea on how I could debug it?

stefanseefeld:
I have been using Boost.Python (including NumPy) on different distros. Notably, we are running unit tests in a trusty container (see https://travis-ci.org/boostorg/python for details) without any issues.

[...]

PR #218, opened 2018-07-11
Add missing return statement in numpy import

### LINKED ISSUE #214: "numpy.core.umath failed to import" with python3
i know #209 has been considered to be an arch-related build problem and thus marked as closed. however, i encountered a very similar import error, with no "endianness error" reported, just `ImportError. numpy.core.umath failed to import`.

base code:
```
#include <boost/python.hpp>
#include <boost/python/numpy.hpp>

void hello() { PySys_WriteStdout("hello\n"); }

BOOST_PYTHON_MODULE(hello) {
  boost::python::numpy::initialize();
  boost::python::def("hello", hello);
}
```

CMakeLists.txt
```
find_package(PythonLibs 3)
find_package(Boost COMPONENTS python3 numpy3)

PYTHON_ADD_MODULE(hello hello.cpp)
target_include_directories(hello PRIVATE ${PYTHON_INCLUDE_DIR})
target_link_libraries(hello Boost::boost Boost::python3 Boost::numpy3 ${PYTHON_LIBRARIES})
```

my build env:
```
os: ubuntu 16.04 x86_64
cc: clang 3.8.0 (NOTE: installed with 'apt install clang')
python: 3.5.2 (NOTE: lib with 'm' suffix, installed with 'apt install python3 python3-dev')
numpy: 1.14.5 (NOTE: installed with 'python3 -m pip install --user numpy')
boost build cmd:
  ./bootstrap.sh --with-toolset=clang --with-python=python3
  ./b2 toolset=clang link=static cxxflags="-std=c++11 -fPIC" variant=release --with-python
```

after some tweaking, i found the following workarounds (modify `libs/python/src/numpy/numpy.cpp`)
1. swap the order of `wrap_import_array()` and `import_ufunc()`
2. **or** replace `wrap_import_array()` with `_import_array()`
2. **or** add `return NULL;` after `import_array()` in `wrap_import_array` implementation for py3

i googled around the web, with no similar result. wierd. hope this workaround would help someone else.

---
another interesting thing is, if i call `wrap_import_array` and `import_ufunc` directly in my own module .c file, as follows:
```
#include <boost/python.hpp>
#include <numpy/arrayobject.h>
#include <numpy/ufuncobject.h>

static void * wrap_import_array() { import_array(); }

void hello() { PySys_WriteStdout("hello\n"); }

BOOST_PYTHON_MODULE(hello) {
  wrap_import_array();
  import_ufunc();

  boost::python::def("hello", hello);
}
```
program crashed with `Illegal instruction` message when `import hello`, stack trace contains no helpful message (i think); however, if i call `np::initialize()` instead, the program would not crash, but here comes the `ImportError. numpy.core.umath failed to import` error.

WanzenBug:
This adds a missing return statement in the python3 specific
import logic of boost.python.numpy.

For python3 wrap_import_array() needs to return a pointer value.
The import_array() macro only returns NULL in case of error. The
missing return statement is UB, so the compiler can assume it does
not happen. This means the compiler can assume the error branch
is always taken, so import_array must always fail.

This fixes #214 and #209

### DIFF
--- a/src/numpy/numpy.cpp
+++ b/src/numpy/numpy.cpp
@@ -19,6 +19,7 @@ static void wrap_import_array()
 static void * wrap_import_array()
 {
   import_array();
+  return NULL;
 }
 #endif
 


stefanseefeld:
So just to make sure I understand your argument: you are saying that, because of the missing return statement, the compiler optimizes away the entire call to `import_array()` ?

WanzenBug:
The argument is:
1. `wrap_import_array()` is expanded by the preprocessor (`import_array()` is just a macro) so it looks like
    ```c++
    static void * wrap_import_array() {
       if (_import_array() < 0) {
          PyErr_Print();
          PyErr_SetString(PyExc_ImportError, "numpy.core.multiarray failed to import"); 
          return NUMPY_IMPORT_ARRAY_RETVAL; 
       }
    }
    ```
    (taken from `/usr/include/numpy/__multiarray_api.h`)
2. No `return <something>` is UB, so the compiler can assume it never happens -> _import_array() must always return < 0.
3. `_import_array()` is also defined in the same `.h` file as `import_array()`, so it can get inlined.
4. `_import_array()` is a long list of `if (condition) { return -1; }`, at the end it has `return 0;`. The compiler can assume this does not happen, as _import_array() must return < 0.
5. (This part I'm not exactly sure about): The compiler knows it cannot reach the last `return 0;` so the last `if (condition)` must always be true. In our case this is exactly the check for endianess.

Issue #171, opened 2017-11-01
feature request: allow changing ndarray WRITABLE flag after ndarray construction


HDembinski:
I want to write a function which creates, fills, and returns an ndarray in C++. The ndarray owns its memory, but should be non-writable for the user. I would therefore like to clear the WRITABLE flag after I finished filling it in my function. I studied the header and I see no way of doing this right now, because flags are not writable after creating an ndarray. For most flags this makes sense, the WRITABLE flag seems to be the only one which you may want to change manually.

Is it possible to add this feature? Best regards!

HDembinski:
Any chance that this will be implemented? Would you review a PR if I prepare one?

stefanseefeld:
Yes, it looks like a simple `ndarray::set_flags()` method is missing, right ?

stefanseefeld:
I'd rather not hand-hold users of the API. So if this is possible in the NumPy C API, I'm inclined to make it accessible in Boost.Python directly. But of course, proper documentation (including warning signs) would be helpful.

HDembinski:
Ok!

HDembinski:
I submitted a PR. Some tests are failing, but this seems unrelated to the new code...

PR #223, opened 2018-08-01
Feature/set_flag

HDembinski:
Addresses #171.

New method `ndarray::set_flag(bitflag, bool on)` is added, which allows to set and clear flags. A corresponding unit test checks whether this works and also that the modification does not change other flags as well.


HDembinski:
Hm, looks like I branched from master and not develop...

PR #224, opened 2018-08-01
add method ndarray::set_flag and corresponding test

HDembinski:
Second attempt, no branching from develop. Addresses #171.


HDembinski:
I don't understand the failing appveyor test. This looks like a configuration issue. I don't see that anything is compiled.

HDembinski:
Appveyor fails with this message:
pip install --disable-pip-version-check --user --upgrade pip
ERROR: To modify pip, please run the following command:
c:\python27\python.exe -m pip install --disable-pip-version-check --user --upgrade pip

This is not caused by this PR.

Issue #228, opened 2018-08-16
error C3861: 'unwind_type': identifier not found


bowie7070:
Hi,

I was using boost.python 1.67.0 quite happily with Visual Studio 15.7.6 and have today upgraded to Visual Studio 15.8.0 and now I get a compilation errors listed below. I am not familiar enough with the boost.python code to establish what has gone wrong so any insights would be appreciated.

1>------ Build started: Project: boost-python-lnlsv, Configuration: Release Win32 ------
1>boost_python_lnlsv.cpp
1>c:\vcpkg\installed\x86-windows\include\boost\python\detail\unwind_type.hpp(37): error C3861: 'unwind_type': identifier not found
1>c:\vcpkg\installed\x86-windows\include\boost\python\detail\unwind_type.hpp(37): note: 'unwind_type': function was not declared in the template definition context and can be found only via argument-dependent lookup in the instantiation context
1>c:\vcpkg\installed\x86-windows\include\boost\python\detail\unwind_type.hpp(59): note: see reference to function template instantiation 'boost::python::converter::detail::unwind_type_id_helper::result_type boost::python::detail::unwind_type_cv<Generator,const U>(const U *,boost::python::detail::const_,Generator *)' being compiled
1>        with
1>        [
1>            Generator=boost::python::converter::detail::unwind_type_id_helper,
1>            U=const LN_LSV::CalibrationParameters
1>        ]
1>c:\vcpkg\installed\x86-windows\include\boost\python\detail\unwind_type.hpp(127): note: see reference to function template instantiation 'boost::python::converter::detail::unwind_type_id_helper::result_type boost::python::detail::unwind_ptr_type<Generator,U>(U *,Generator *)' being compiled
1>        with
1>        [
1>            Generator=boost::python::converter::detail::unwind_type_id_helper,
1>            U=const LN_LSV::CalibrationParameters
1>        ]
1>c:\vcpkg\installed\x86-windows\include\boost\python\detail\unwind_type.hpp(165): note: see reference to function template instantiation 'boost::python::converter::detail::unwind_type_id_helper::result_type boost::python::detail::unwind_helper2<2>::execute<Generator,const LN_LSV::CalibrationParameters>(U &(__cdecl *)(void),Generator *)' being compiled
1>        with
1>        [
1>            Generator=boost::python::converter::detail::unwind_type_id_helper,
1>            U=const LN_LSV::CalibrationParameters
1>        ]
[...]

PR #229, opened 2018-08-16
Enable forward declaration of unwind_type() in msvc14.15 and later.

bowie7070:
Name lookup in msvc has changed between 14.14 and 14.15 making it consistent with other compilers. Forward declaration of unwind_type() is now required as it is for other compilers.

Resolves compilation errors identified in:

https://github.com/boostorg/python/issues/228


stefanseefeld:
Thanks !

PR #237, opened 2018-10-12
Add empty stage/install targets when Python is not configured

pdimov:
Since the previously added boost-install is in `if [ python.configured ]`, this `else` adds stage/install targets (that issue the preexisting warning message) in case Python hasn't been configured.


pdimov:
Done.

pdimov:
Can we please have this, along with https://github.com/boostorg/python/pull/236, merged to master?

Issue #240, opened 2018-11-06
Getting a boost python function's docstring causes Segfault on win32/MSVC builds.


ricrogz:
This bug triggers when accessing the `__doc__` property of a boost python wrapped function. It was detected on importing the `rdkit.Chem.Lipinski` module of the RDKit (see line 95 [here](https://github.com/rdkit/rdkit/blob/master/rdkit/Chem/Lipinski.py); `rdMolDescriptors` is boost python wrapped). The problem was detected on Windows 7 x64, having built both boost and RDKit with VS2017, and has been reproduced on a different machine (also Windows 7 x64) in the conda environment. `gcc` builds on linux do not seem to exhibit this bug.

The segfault apparently happens because, `s.basename` for the function's return type has vanished and contains a `nullptr` when evaluated inside `function_doc_signature_generator::py_type_str` ([https://github.com/boostorg/python/blob/develop/src/object/function_doc_signature.cpp](https://github.com/boostorg/python/blob/develop/src/object/function_doc_signature.cpp), line 119).

mnordie:
Possible duplicate of #207?
We've used PR #208 to correct this issue. It's not in the 1.69 beta though.

ricrogz:
Hi @mnordie, actually, yes, it seems to be a duplicate. I tried #208, and it solves the problem. Thanks, and sorry for the trouble!

PR #241, opened 2018-11-06
Fix function_doc_signature_generator::py_type_str causing segfault (issue #240)

ricrogz:
Issue #240 is apparently caused by `s.basename` vanishing before being evaluated, and thus comparing a `nullpointer`.

Storing the return type's name in a `static std::string` prevents the name from vanishing before being checked.


ricrogz:
Closing, since the issue is a duplicate of #207, and PR #208 solves it.

PR #249, opened 2019-02-13
clear python exception after expected attribute lookup failure

orivej:
Python 3.7.0 asserts that attribute lookup functions are called without outstanding exceptions:
https://github.com/python/cpython/blob/v3.7.0a2/Objects/typeobject.c#L3037
Motivation: https://bugs.python.org/issue34068#msg321262
Therefore the error set by the first PyObject_GetItem should be cleared before calling PyObject_GetAttrString.

### DIFF
--- a/src/object/function.cpp
+++ b/src/object/function.cpp
@@ -444,7 +444,9 @@ void function::add_to_namespace(
         if (dict == 0)
             throw_error_already_set();
 
+        assert(!PyErr_Occurred());
         handle<> existing(allow_null(::PyObject_GetItem(dict.get(), name.ptr())));
+        PyErr_Clear();
         
         if (existing)
         {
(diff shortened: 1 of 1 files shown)


src/object/function.cpp
```diff
@@ -485,16 +487,15 @@ void function::add_to_namespace(
         if (new_func->name().is_none())
             new_func->m_name = name;
 
+        assert(!PyErr_Occurred());
```
stefanseefeld: I don't quite understand: this code here is conditional on `existing` being non-null. So the only way to be here is with the previous `PyObject_GetItem()` call not having failed. So why the `PyErr_Clear()` call between that and here ?

orivej: Why is it conditional on `existing`? It is outside `if (existing)`.

stefanseefeld: oups, sorry, I misread (the diff view collapsed a few important lines here :-) ). OK, makes sense, then.

PR #251, opened 2019-03-03
 added support for 'set'

FadyEssam:


include/boost/python/set.hpp
```diff
@@ -0,0 +1,96 @@
+#ifndef SET_BOOST_PYTHON_HH
+#define SET_BOOST_PYTHON_HH
+
+#include <boost/python.hpp>
+#include <boost/python/ssize_t.hpp>
+
+namespace boost {
+	namespace python {
```
stefanseefeld: please reformat the patch to use two spaces as indentation, rather than eight.

stefanseefeld:
Given the similarities with other Python types, perhaps you can guess what build logic needs to be added ? (Have a look at `build/Jamfile`).
Similarly, perhaps you could clone some of the logic from `test/dict.*` to come up with a test for `set` ?

include/boost/python/set.hpp
FadyEssam: I think all the mentioned issues are resolved now

stefanseefeld:
There are still a few pieces of build infrastructure missing:
* add the new source file to https://github.com/boostorg/python/blob/develop/src/fabscript,,,
* ...as well as https://github.com/boostorg/python/blob/develop/build/Jamfile
* Add the test to https://github.com/boostorg/python/blob/develop/test/fabscript

FadyEssam:
done, Thanks for the guidance

src/set.cpp
```diff
@@ -0,0 +1,83 @@
...
+	  void set_base::discard(object_cref x)
+	  {
+		if (PyAnySet_CheckExact(this->ptr()))
+		{
+		  if (PySet_Discard(this->ptr(), x.ptr()) == -1)
+			throw_error_already_set();
+		}
+		else
+		{
+		  this->attr("discrad")(x);
```
lightmare: there's a typo here

PR #252, opened 2019-03-18
Fix -Wregister error on Clang in C++17 mode

Kojoley:
Suppresses the warning on GCC, Clang, and MSVC.

Probably it is worth cherry-picking into the master as even the release managers had the trouble with Python 2 headers.

Playground: https://godbolt.org/z/ioqTlS


Kojoley:
Ping

stefanseefeld:
Thanks !

Issue #254, opened 2019-03-22
CI/faber seems to not run some of the tests


Kojoley:
Some tests are failing with b2 while here on CI they are fine:
  * `injected`
  * `boost_shared_ptr`
  * `shared_ptr`
  * `auto_ptr`
  * `operators`
  * `implicit`
  * `data_members`
  * `extract`
  * `numpy~dtype`

You can look at them on regression matrix or here https://travis-ci.com/Kojoley/python/jobs/187163937

PR #257, opened 2019-03-27
Fixed calling_conventions tests on MinGW

Kojoley:
On MinGW calling conversion are actually macros that are expanded into a GCC attribute (e.g. `__fastcall` expands into `__attribute__((__fastcall__))`).

### DIFF
--- a/test/calling_conventions.cpp
+++ b/test/calling_conventions.cpp
@@ -31,15 +31,15 @@ using namespace boost::python;
 
 #define TEST_DECLARE_FUNCTIONS
 
-#define TESTED_CALLING_CONVENTION __cdecl
+#define TESTED_CALLING_CONVENTION cdecl
 #include "calling_conventions.cpp"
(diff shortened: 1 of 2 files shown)


Kojoley:
Ping

stefanseefeld:
Sorry for my silence. Me not being a Windows developer I feel a bit uncomfortable reviewing this. Do you think it would be possible to add a MinGW (cross-)compiler to the mix of compilers we use on Travis-CI or AppVeyor ? That would help prove that things work as intended.

Kojoley:
I seen other boost projects use mingw on appveyor, I could have added it here, but I do know nothing about faber (and it seems to not run module tests https://github.com/boostorg/python/issues/254).

stefanseefeld:
OK, fair point. I'll thus try to add MinGW to the mix, then we should be able to see whether this PR fixes one (or more) failures.

PR #259, opened 2019-03-30
Fixed module tests fail

Kojoley:
Since recently lightweight_test introduced a sanitation to ensure that `boost::report_errors()` is called, and it does not in module tests.

The problem could be fixed by exporting the function and calling it from these python tests, but as it already done in other module tests I just turned those usages lightweight_test to a regular assertation.

ref #254

### DIFF
--- a/test/test_class.hpp
+++ b/test/test_class.hpp
@@ -4,17 +4,17 @@
 // http://www.boost.org/LICENSE_1_0.txt)
 #ifndef TEST_CLASS_DWA2002326_HPP
 # define TEST_CLASS_DWA2002326_HPP
-# include <boost/detail/lightweight_test.hpp>
+# include <boost/assert.hpp>
 
 template <int n = 0>
 struct test_class
 {
     explicit test_class(int x) : x(x), magic(7654321 + n) { ++counter; }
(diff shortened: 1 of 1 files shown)


stefanseefeld:
I don't think `BOOST_TEST` is a device meant to be used for testing. Notably, it will empty (and the enclosed expression won't be evaluated) if `NDEBUG` is defined, just like the standard `assert` macro.

Kojoley:
Then fix it like you think it should be.

Also grep the tests for `BOOST_ASSERT`.

The cases closely related to this PR:
https://github.com/boostorg/python/blob/ac62db1cf17c0af44dc2b2117f9ddeee327e5e7c/test/virtual_functions.cpp#L15-L34
https://github.com/boostorg/python/blob/ac62db1cf17c0af44dc2b2117f9ddeee327e5e7c/test/back_reference.cpp#L24-L39
https://github.com/boostorg/python/blob/ac62db1cf17c0af44dc2b2117f9ddeee327e5e7c/test/callbacks.cpp#L31-L46

stefanseefeld:
OK, fair enough. I'll merge this in an attempt to "harmonize" our testing logic. Then we can open a new request to substitute `boost_assert` by something more robust and suitable for use with testing.

PR #264, opened 2019-04-17
Only boost-install boost_numpy when it's being built

pdimov:
The library `boost_numpy` is conditioned on `[ python.numpy ]`, so only `boost-install` it when this condition holds.

This also moves the libraries outside `if [ python.configured ]` (as was the case before) since they are already guarded with `[ cond ]`.

### DIFF
--- a/build/Jamfile
+++ b/build/Jamfile
@@ -30,8 +30,6 @@ else
         ;
 }
 
-if [ python.configured ]
-{
 project boost/python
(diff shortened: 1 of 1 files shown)


stefanseefeld:
Thanks ! (Feel free to merge this right away without waiting for CI builds to finish. They don't use `b2`, so wouldn't be able to detect any issues anyhow. :-)

PR #291, opened 2020-03-02
Use python macros to permit alternate pythons (start pypy)

jcpunk:
Resolves https://svn.boost.org/trac10/ticket/4125

This change uses the official python macros for object access.  This should have no impact on existing CPython usage, but will permit alternate pythons (such as pypy) to utilize boost.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,20 +28,26 @@ jobs:
       env: CXX=clang++ PYTHON=python3 CXXFLAGS=-std=c++98
     - os: linux
       env: CXX=clang++ PYTHON=python3 CXXFLAGS=-std=c++11
+    - os: linux
+      env: CXX=g++ PYTHON=pypy3 CXXFLAGS=-std=c++11
     - os: osx
       env: CXX=clang++ PYTHON=python CXXFLAGS=-std=c++11
     - env: PYTHON=python DOC=1
   allow_failures:
+    - os: linux
+      env: CXX=g++ PYTHON=pypy3 CXXFLAGS=-std=c++11
     - os: osx
 
 addons:
   apt:
     sources:
     - ubuntu-toolchain-r-test
+    - pypy
     packages:
     - gcc
     - g++
     - clang
+    - pypy3-dev
     - python3-pip
     - python-numpy
     - python-sphinx
--- a/src/exec.cpp
+++ b/src/exec.cpp
@@ -104,14 +104,22 @@ object BOOST_PYTHON_DECL exec_file(char const *filename, object global, object l
   if (local.is_none()) local = global;
   // should be 'char const *' but older python versions don't use 'const' yet.
   char *f = const_cast<char *>(filename);
-  // Let python open the file to avoid potential binary incompatibilities.
-#if PY_VERSION_HEX >= 0x03040000
-  FILE *fs = _Py_fopen(f, "r");
-#elif PY_VERSION_HEX >= 0x03000000
+#if PY_VERSION_HEX >= 0x03010000
+  // Let python manage any UTF bits to avoid potential incompatibilities.
   PyObject *fo = Py_BuildValue("s", f);
-  FILE *fs = _Py_fopen(fo, "r");
+  PyObject *fb = Py_None;
+  PyUnicode_FSConverter(fo, &fb);
+  f = PyBytes_AsString(fb);
+  FILE *fs = fopen(f, "r");
+  Py_DECREF(fo);
+  Py_DECREF(fb);
+#elif PY_VERSION_HEX >= 0x03000000
+  // Let python open the file to avoid potential binary incompatibilities.
+  PyObject *fo = Py_BuildValue("s", f);
+  FILE *fs = _Py_fopen(fo, "r"); // Private CPython API
   Py_DECREF(fo);
 #else
+  // Let python open the file to avoid potential binary incompatibilities.
   PyObject *pyfile = PyFile_FromString(f, const_cast<char*>("r"));
   if (!pyfile) throw std::invalid_argument(std::string(f) + " : no such file");
   python::handle<> file(pyfile);
--- a/src/wrapper.cpp
+++ b/src/wrapper.cpp
@@ -25,7 +25,7 @@ namespace detail
             
               if (
                   PyMethod_Check(m.get())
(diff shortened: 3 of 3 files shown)


mattip:
While we are at it, could you replace the use of the CPython private `_Py_fopen` with the standard `fopen`?

jcpunk:
I'm not sure how that would interact with Unicode translations....  I'm worried (perhaps unnecessarily) that just aliasing `_Py_fopen` to `fopen` on PyPy might hide some sort of character encoding translation bug type thing.  I know almost nothing on the character encoding front....

Looking through the commit history, I'm trying to understand the rational between https://github.com/boostorg/python/blob/develop/src/exec.cpp#L108 and https://github.com/boostorg/python/blob/develop/src/exec.cpp#L110.

I suppose I've a few questions on that front:

- What is the oldest version of CPython (2 and 3) this code should support?
- What is the oldest version of PyPy we would want to support?

mattip:
The CI run I looked at failed with 
```
Error: /usr/local/lib/python2.7/dist-packages/faber/tools/python.py:98: \
'python' object has no attribute 'lib'
```

mattip:
I think [this line](https://github.com/boostorg/python/blob/develop/.appveyor.yml#L67) in appveyor' setup must be changed to `python -m pip install ...`

jcpunk:
The AppVeyor bits seem to have a PR https://github.com/boostorg/python/pull/294

jcpunk:
Looks like Python 3.10 will complain about the non-public symbols.

I've reworked the patches a bit to make pypy optional in Travis CI for now.

stefanseefeld:
Hi there, sorry for the high latency. I have no experience with PyPy, and thus will require a bit of help in adding support for it to `boost.python`. Given that I'm using `faber` to do the build, I'll start by adding the required `PyPy` support there (https://github.com/stefanseefeld/faber/issues/21).

jcpunk:
Sounds good!

Commits 1 and 2 from this branch will probably be required for cPython 3.10 final.

.travis.yml
```diff
@@ -28,20 +28,32 @@ jobs:
       env: CXX=clang++ PYTHON=python3 CXXFLAGS=-std=c++98
     - os: linux
       env: CXX=clang++ PYTHON=python3 CXXFLAGS=-std=c++11
+    - os: linux
+      env: CXX=g++ PYTHON=pypy CXXFLAGS=-std=c++11
+    - os: linux
+      env: CXX=clang++ PYTHON=pypy CXXFLAGS=-std=c++98
```
stefanseefeld, line 34: I'm not convinced such extensive coverage is needed. I'd only add two new build variants to cover pypy and pypy3 with C++11.

.travis.yml
```diff
@@ -28,20 +28,32 @@ jobs:
...
+      env: CXX=g++ PYTHON=pypy3 CXXFLAGS=-std=c++11
+    - os: linux
+      env: CXX=clang++ PYTHON=pypy3 CXXFLAGS=-std=c++98
     - os: osx
       env: CXX=clang++ PYTHON=python CXXFLAGS=-std=c++11
     - env: PYTHON=python DOC=1
   allow_failures:
     - os: osx
 
+
```
stefanseefeld, line 45: extra newline

.travis.yml
jcpunk: I'll drop those and re-push.

jcpunk:
In theory I've addressed your requests and Travis will validate that shortly..... in practice.... well we'll see....

stefanseefeld:
Thanks, the PR looks good. I'll approve once the builds complete successfully.

stefanseefeld:
I do see a bunch of test failures, but suspect that they are at least partially caused by my recent changes to `Faber`. I'll investigate.

stefanseefeld:
OK, so I have fixed the `Faber` tool's failures. Now the only failures are with the newly added `PyPy` backends.

jcpunk:
The remaining errors generated on the pypy compile seem to be:

```
include/boost/python/detail/is_auto_ptr.hpp:17:40: warning: ‘template<class> class std::auto_ptr’ is deprecated

 BOOST_PYTHON_IS_XXX_DEF(auto_ptr, std::auto_ptr, 1)
                                        ^
```

It looks like C++11 added `unique_ptr` to use instead, but that isn't available in older C++ versions.[1]

I'm pretty sure I'm not qualified to assess the risks in modifying this.....


[1]
https://en.wikipedia.org/wiki/Auto_ptr
https://en.wikipedia.org/wiki/Smart_pointer#unique_ptr

stefanseefeld:
I don't think `std::auto_ptr` is causing the failures. The above is just a warning, and it's present with the CPython variants as well. (You can disable it by adding `cxxflags=-W-no-deprecated-warnings`)

stefanseefeld:
There is, however, this error:
```
./src/object/function.cpp: In static member function ‘static void boost::python::objects::function::add_to_namespace(const boost::python::api::object&, const char*, const boost::python::api::object&, const char*)’:
./src/object/function.cpp:436:60: error: ‘struct PyClassObject’ has no member named ‘cl_dict’
             dict = handle<>(borrowed(((PyClassObject*)ns)->cl_dict));
```

stefanseefeld:
...and the `pypy3` variant has a bunch of test failures with
```
RPython traceback:
  File "pypy_module_cpyext.c", line 57899, in wrapper_second_level__star_2_36
  File "pypy_module_cpyext_1.c", line 57139, in PyType_IsSubtype
  File "pypy_module_cpyext_1.c", line 35912, in from_ref
Fatal RPython error: AssertionError
Aborted (core dumped)
```

jcpunk:
With python 2 being EOL, I'm thinking we don't need to support `pypy2`.  I'm not sure which users would want a "new boost" but an old pypy.

I'll have to look into the `pypy3` failures more closely.  It looks like a the reference counting bits may need some review...

For now I've flipped `pypy3` into `allowed failure` as commits 1 and 2 from here are pre-reqs for pypy and look like they'll be required for python 3.10.

stefanseefeld:
OK, fair enough.

jcpunk:
With the builds passing for CPython here, I'd say this part is complete.  The `pypy3` errors look like they'll require some more invasive work.

stefanseefeld:
Thanks for the PR !

PR #293, opened 2020-03-19
Ensure all doctests run

tkieft:
905b34210b9ae2310c61b52584313e33b7f131ca added Python 3 compatibility for the tests. However, due to the behavior of doctest, this silently disabled some of the tests. For doctest to pick up module docstrings, the docstring has to be the first statement of the module. You can see that the tests are failing to execute if you add `verbose=True` to the `doctest.testmod()` call.

This PR fixes this error for all 12 modules where there was a non-docstring as the first statement, and then adds minor fixups to the tests to make sure they execute on both Python 2 and Python 3.

### DIFF
--- a/test/dict.cpp
+++ b/test/dict.cpp
@@ -21,11 +21,13 @@ object new_dict()
 object data_dict()
 {
     dict tmp1;
-    tmp1["key1"] = "value1";
 
     dict tmp2;
     tmp2["key2"] = "value2";
     tmp1[1] = tmp2;
+
+    tmp1["key1"] = "value1";
+
     return tmp1;
 }
 
@@ -60,22 +62,20 @@ void work_with_dict(dict data1, dict data2)
 void test_templates(object print)
 {
     std::string key = "key";
-    
+
     dict tmp;
-    tmp[1] = "a test string";
-    print(tmp.get(1));
-    //print(tmp[1]);
     tmp[1.5] = 13;
     print(tmp.get(1.5));
+    tmp[1] = "a test string";
+    print(tmp.get(1));
     print(tmp.get(44));
     print(tmp);
     print(tmp.get(2,"default"));
     print(tmp.setdefault(3,"default"));
 
     BOOST_ASSERT(!tmp.has_key(key));
-    //print(tmp[3]);
 }
-    
+
 BOOST_PYTHON_MODULE(dict_ext)
 {
     def("new_dict", new_dict);
--- a/test/fabscript
+++ b/test/fabscript
@@ -21,14 +21,14 @@ src = module('..src')
 python_libs=python.instance().libs
 features |= runpath(src.bpl.path, base='')
 
-def extension_test(name, ext=[], script=None, np=False,
+def extension_test(name, exts=[], script=None, numpy=False,
                    features=features, condition=None):
     """Create a Python extension test `name`.
     Arguments:
       * name: the name of the test.
-      * ext: extensions to be compiled, <name> if none are given.
+      * exts: extensions to be compiled, <name> if none are given.
       * script: the test script to execute, <name>.py if none is given.
-      * np: if true, add boost_numpy to sources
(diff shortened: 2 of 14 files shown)


test/dict.cpp
```diff
@@ -21,11 +21,13 @@ object new_dict()
...
 {
     dict tmp1;
-    tmp1["key1"] = "value1";
 
     dict tmp2;
     tmp2["key2"] = "value2";
     tmp1[1] = tmp2;
+
+    tmp1["key1"] = "value1";
+
```
stefanseefeld: Just for avoidance of doubt: this reflects the fact that in current versions of Python the order reported (notably when printing a dict) corresponds to the insertion order, yes ?

tkieft: Yes! If we want to print out the full dictionary, we have to make sure the output will be the same across Python versions. I'm not sure what ordering Python 2 is using (by hash, perhaps), but Python 3 is using insertion order.

stefanseefeld: Right. So if ever we find the order to vary (for older Python versions, where the order isn't mandated yet), we need to rethink how to print the dict...

tkieft: I think I tested each case on my local copy of Python2, and the order seemed fairly stable. Hopefully the same shows up in the CI.

tkieft:
Looking at the test failures, let me update :)

tkieft:
I think one of the errors (`test.map_indexing_suite`) is due to faber not building the native extension correctly. Let me try submitting a separate PR to update Travis to use faber-0.3 (which works on my machine) and see if that will fix things.

test/test_builtin_converters.py
```diff
@@ -136,9 +134,6 @@
 >>> print(rewrap_value_wstring(u'yo, wassup?'))
 yo, wassup?
 
->>> print(rewrap_value_wstring(u'\U0001f4a9'))
```
tkieft: I had to remove this test; there was no way to get it to work on Python 2 & 3 simultaneously. Luckily the `rewrap_value_wstring` function is still tested above.

tkieft:
@stefanseefeld I think I've fixed all the test errors except for `map_indexing_suite`. The problem there is that faber is building the extension with the wrong name (`a_map_indexing_suite` instead of `map_indexing_suite_ext`).  This doesn't happen on my machine, and I can't figure out why it's happening. Can you help debug?

stefanseefeld:
Yes, I'll look into the faber issue. (I don't have a Windows machine, and running things on a Windows VM is a bit painful.)

tkieft:
Actually, hold off for a moment - I think I might have fixed it. Just pushed a new build to check 20 minutes ago :)

test/fabscript
```diff
@@ -21,32 +21,32 @@ src = module('..src')
...
-            n = e if e != name else e + '_ext'
-            s = [e + '.cpp']
+    libs = [src.bnl, src.bpl] if numpy else [src.bpl]
+    for ext in exts or [name]:
+        if type(ext) is str:  # build from a single source file
+            ext_name = ext if ext != name else ext + '_ext'
+            sources = [ext + '.cpp']
         else:  # build from a list of source files
-            n = e[0] if e[0] != name else e[0] + '_ext'
-            s = [n + '.cpp' for n in e]
```
tkieft: In Python 2, this `n` will leak out of the generator and trample over the previous value.

stefanseefeld: Great finding ! Good that Python 2 has reached its end of life. I'm about to remove Python 2 support from Faber, too...

tkieft:
OK, Travis succeeded, I think this is ready for a closer look! The AppVeyor failures it looks like you're already aware of, and are unrelated to this diff.

tkieft:
@stefanseefeld Thoughts on merging this? Are you trying to find time to get the Windows CI running?

tkieft:
@stefanseefeld - would really love to get this in!

PR #324, opened 2020-10-16
Convert get/set/size to PEP-384 style

jcpunk:
Ultimately, for PEP-384 (CPython 3.2+) someone who understands the output needs to run `abitype.py`[1] against `src/object/enum.cpp` and `src/object/class.cpp`.  The slot/spec layouts[2] generated by that code are pretty close to right, but need someone more fluent in this than I....

This gets the other helpers in place for staying to the stricter ABI.

[1]
https://github.com/python/cpython/blob/master/Tools/scripts/abitype.py

[2]
https://pythoncapi.readthedocs.io/type_object.html#pytype-fromspec


knoepfel:
@jcpunk, do you have reason to suspect the output of `abitype.py` on those files is wrong?

jcpunk:
I'm really unsure what the changes mean for the overall usage and behaviour of the library.  To be honest, I'm not much of a C++ programmer and can't meaningfully assess that output.

stefanseefeld:
I'm not entirely sure I understand the purpose of this PR. As far as I can see, all it does is to replace non-checking functions by their checking counter parts, thereby making the code overall slower. (Example: `PyTuple_GET_SIZE` is replaced by `PyTuple_Size`, `PyTuple_GET_ITEM` is replaced by `PyTuple_GetItem`, etc.)
Am I missing something ?

jcpunk:
As I understand it, this PR (switching to the checking versions) is a pre-req for the PEP-384 `abitype.py` conversion to the stable C API.

PR #328, opened 2020-11-14
Fix Python 3.10 (PEP-620) incompatibility.

stefanseefeld:


include/boost/python/detail/wrap_python.hpp
```diff
@@ -227,7 +227,11 @@ typedef int pid_t;
 
 # define PyVarObject_HEAD_INIT(type, size) \
         PyObject_HEAD_INIT(type) size,
+#endif
 
+#if PY_VERSION_HEX < 0x030900A4
+#  define Py_SET_TYPE(obj, type) ((Py_TYPE(obj) = (type)), (void)0)
+#  define Py_SET_SIZE(obj, size) ((Py_SIZE(obj) = (size)), (void)0)
```
vstinner, line 234: I wrote PR #330 which uses pythoncapi_compat.h to not have to define these compatibility macros manually.

PR #329, opened 2020-12-09
Use Py_TYPE() rather than using PyObject.ob_type

vstinner:
Py_TYPE() is available since Python 2.6.

### DIFF
--- a/src/converter/builtin_converters.cpp
+++ b/src/converter/builtin_converters.cpp
@@ -154,7 +154,7 @@ namespace
   {
       static unaryfunc* get_slot(PyObject* obj)
       {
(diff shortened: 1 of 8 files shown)


vstinner:
PR created by the "Py_TYPE" operation of the upgrade_pythoncapi.py script of https://github.com/pythoncapi/pythoncapi_compat

vstinner:
See also https://github.com/boostorg/python/pull/329

vstinner:
Ah, the change was wrong on tests: x.get() should be used to get the `PyObject*`.

vstinner:
No feedback in 2 years. I just close my PR.

PR #330, opened 2020-12-09
Use Py_SET_SIZE() function

vstinner:
Replace "Py_TYPE(obj)=type;" with "Py_SET_SIZE(obj, type);". The
Py_SET_SIZE() was added to Python 3.9: copy pythoncapi_compat.h
header to src/object/. The file comes from the project:
https://github.com/pythoncapi/pythoncapi_compat

On Python 3.10, Py_SET_SIZE() should now be used rather than using
Py_TYPE() as an l-value:
https://docs.python.org/dev/c-api/structures.html#c.Py_TYPE

Context: https://bugs.python.org/issue39573

### DIFF
--- a/src/object/class.cpp
+++ b/src/object/class.cpp
@@ -1,3 +1,5 @@
+#include "pythoncapi_compat.h"
+
 // Copyright David Abrahams 2001.
 // Distributed under the Boost Software License, Version 1.0. (See
 // accompanying file LICENSE_1_0.txt or copy at
@@ -208,7 +210,7 @@ namespace objects
   {
       if (static_data_object.tp_dict == 0)
       {
-          Py_TYPE(&static_data_object) = &PyType_Type;
+          Py_SET_TYPE(&static_data_object, &PyType_Type);
           static_data_object.tp_base = &PyProperty_Type;
           if (PyType_Ready(&static_data_object))
               return 0;
@@ -316,7 +318,7 @@ namespace objects
   {
       if (class_metatype_object.tp_dict == 0)
       {
-          Py_TYPE(&class_metatype_object) = &PyType_Type;
+          Py_SET_TYPE(&class_metatype_object, &PyType_Type);
           class_metatype_object.tp_base = &PyType_Type;
           if (PyType_Ready(&class_metatype_object))
               return type_handle();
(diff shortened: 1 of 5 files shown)


vstinner:
See also https://github.com/boostorg/python/pull/329

stefanseefeld:
This looks like a duplicate of https://github.com/boostorg/python/pull/328, doesn't it ?

vstinner:
> This looks like a duplicate of #328, doesn't it ?

PR #328 replaces `obj->ob_type` with `Py_TYPE()`. It doesn't need pythoncapi_compat.h, since Py_TYPE() exists since Python 2.6.

stefanseefeld:
Aren't you referring to https://github.com/boostorg/python/pull/329 instead ?

vstinner:
> This looks like a duplicate of #328, doesn't it ?

Oh wait, ignore my previous comment. I wrote 2 PRs and I didn't notice that you're a talking about a 3rd PR.

Yes, #328 is the same fix, but it doesn't use pythoncapi_compat.h. I propose to use it to avoid having to reimplemented the compatibility layer for old Python versions:

```
#if PY_VERSION_HEX < 0x030900A4
#  define Py_SET_TYPE(obj, type) ((Py_TYPE(obj) = (type)), (void)0)
#  define Py_SET_SIZE(obj, size) ((Py_SIZE(obj) = (size)), (void)0)
#endif
```

pythoncapi_compat.h uses static inline functions rather than macros.

vstinner:
Fedora uses a downstream patch to fix issue: https://src.fedoraproject.org/rpms/boost/blob/master/f/boost-1.73-python3.10.patch Copy of the boost specfile:

```
# https://bugzilla.redhat.com/show_bug.cgi?id=1896382
# https://github.com/boostorg/python/issues/325
Patch93: boost-1.73-python3.10.patch
```

vstinner:
> APPVEYOR_BUILD_WORKER_IMAGE=Visual Studio 2019

This job failed because it failed to locate the C++ compiler, whereas the two other AppVeyor jobs passed:

```
set FARGS=--log=summary --log=output --log=actions --log=commands --with-boost-include=%BOOST_PREFIX% target.arch=%ARCH% cxx.name=msvc cxx.version=%MSVC%
faber %FARGS% config || type config.log
Error: C:\Python36-x64\lib\site-packages\faber\tools\cxx.py:64: no C++ compiler found matching <set cxx.name=msvc cxx.version=16.7.4 target.arch=x86_64>.
faber %FARGS% -j8
Error: C:\Python36-x64\lib\site-packages\faber\tools\cxx.py:64: no C++ compiler found matching <set cxx.name=msvc cxx.version=16.7.4 target.arch=x86_64>.
Command exited with code 1
```

stefanseefeld:
Closed as obsolete.

PR #344, opened 2021-01-06
Fix compatibility with Python 3.10.0a4: fopen

vstinner:
Replace private _Py_fopen() with public fopen(): private _Py_fopen()
function was removed in 3.10.0a4:
https://bugs.python.org/issue32381


vstinner:
Fedora downstream issue: https://bugzilla.redhat.com/show_bug.cgi?id=1912903

vstinner:
If Boost needs a public C API for https://www.python.org/dev/peps/pep-0446/ please ask for it :-) But it would be a new Python feature. In general, third party projects should not rely on private functions, or be prepared for breakage at new Python version.

stefanseefeld:
Thanks !

PR #349, opened 2021-01-26
delete #define _DEBUG

shelltdf:
this is bug


Osyotr:
This is not a bug. `_DEBUG` was temporarily undefined here. https://github.com/boostorg/python/blob/a6dbc37254689edd9efbd8ca4e4771baaf801447/include/boost/python/detail/wrap_python.hpp#L45

PR #351, opened 2021-02-03
Fix error handling inside BOOST_PYTHON_MODULE()

### LINKED ISSUE #280: Handling errors inside BOOST_PYTHON_MODULE()
I have a module that may fail to initialize due to missing dependencies. So I do
```c++
BOOST_PYTHON_MODULE(mymodule)
{
...
    if (missing_dependency) {
        throw std::exception("Missing dependency");
    }
...
}
```
The exception is correctly caught and `PyErr_SetString(PyExc_RuntimeError, x.what())` is called, but python reports
```
    import mymodule
SystemError: initialization of mymodule raised unreported exception
```
Why `unreported exception`? My platform is Windows 10/Python 3.7.4 x64/MSVC 2017 if that matters. python_boost version is 1.69.0.

db4:
Closes #280


stefanseefeld:
Thanks for the patch !

PR #362, opened 2021-05-24
Use the /python//numpy target instead of [ numpy.include ] (fixes #361)

### LINKED ISSUE #361: Boost.NumPy doesn't build when more than one Python version is configured in user-config
When user-config.jam has more than one version of Python configured, e.g.
```
using python : 2.7 : C:/Python27 ;
using python : 3.9 : C:/Python39 ;
```
trying to build Python (with e.g. `b2 --with-python python=2.7`) fails, because Boost.NumPy is built against the (correct) 2.7 version of Python, but against the 3.9 headers of NumPy:
```
cl /Zm800 -nologo "libs\python\src\numpy\ufunc.cpp" -Fo"bin.v2\libs\python\build\msvc-14.2\debug\address-model-32\link-static\python-2.7\threading-multi\numpy\ufunc.obj"    -TP /wd4675 /EHs /GR /Zc:throwingNew /Z7 /Od /Ob0 /W3 /MDd /Zc:forScope /Zc:wchar_t /Zc:inline -c -DBOOST_ALL_NO_LIB=1 -DBOOST_NUMPY_SOURCE -DBOOST_NUMPY_STATIC_LIB -DBOOST_PYTHON_STATIC_LIB "-I." "-IC:\Python27\Include" "-IC:\Python39\lib\site-packages\numpy\core\include"
```
The reason for that is that the Python configuration in b2's `python.jam` stores the NumPy include path in a variable `.numpy-include`, which can only hold one value.

https://github.com/boostorg/build/blob/916afd3876bc0d5b706a63f9d09ba66ff69e7d66/src/tools/python.jam#L1087-L1090
https://github.com/boostorg/build/blob/916afd3876bc0d5b706a63f9d09ba66ff69e7d66/src/tools/python.jam#L858

This can in principle be fixed by declaring a target for NumPy's include path, instead of a variable, and then linking to it instead of using `<include>$(numpy-include)` in build/Jamfile.

https://github.com/boostorg/python/blob/2a82afdf6d30c28eb1775f3b8041d988f6f04598/build/Jamfile#L120

pdimov:
This fixes issue #361 by using the `/python//numpy` target. It however requires the corresponding Boost.Build PR https://github.com/boostorg/build/pull/726 to be merged first.

### DIFF
--- a/build/Jamfile
+++ b/build/Jamfile
@@ -117,7 +117,7 @@ lib boost_numpy
         <define>BOOST_NUMPY_SOURCE
         [ cond [ python.numpy ] : <library>/python//python_for_extensions ]
         [ unless [ python.numpy ] : <build>no ]
-        <include>$(numpy-include)
+        <library>/python//numpy
         <library>boost_python
         <python-debugging>on:<define>BOOST_DEBUG_PYTHON
         -<tag>@$(BOOST_JAMROOT_MODULE)%$(BOOST_JAMROOT_MODULE).tag


SoapGentoo:
I've verified that this patch works with multiple python implementations, please merge it.

thesamesam:
As per my [comment](https://github.com/boostorg/build/pull/726#issuecomment-847355490) on the linked PR for the other side in Build, this resolves my issue.

stefanseefeld:
@pdimov , can you please confirm that it's now safe to (re-)merge this PR to devel (so I can merge everything to master before the upcoming deadline) ?

pdimov:
I can see the change on b2's master branch (https://github.com/boostorg/build/blob/d52f1993d906aa3fd3649a8dd5134402794bf86e/src/tools/python.jam#L1055-L1057), so this should now work. (cc @grafikrobot)

stefanseefeld:
Thanks, I have re-added your commit, and will merge to master before Wednesday.

pdimov:
Reminder that it's Wednesday. :-)

SoapGentoo:
Can we get this merged before 1.77 hits?

Issue #371, opened 2021-08-10
exec_file fails completely on Python 3.1 and above


hajokirchhoff:
When PY_VERSION_HEX >= 0x03010000, boost::python::exec_file passes an invalid string to Py_RunFile.

The problem is that char* f points to a temporary buffer returned by the UTF conversion routine PyBytes_AsString, but this buffer is deallocated via Py_DECREF before f gets passed to Py_RunFile.

I will create a pull request with a fix shortly.

PR #372, opened 2021-08-10
Fix for #239 and #371 - fixes two bugs in `exec_file`

### LINKED ISSUE #239: exec_file does not close file
When I run exec_file in a loop, I get an error on the 509th run. Using process monitor I can see that after each call to exec_file, CloseFile is not being called.

I believe fe24ab9dd5440562e27422cd38f7de03356bfd16 fixed it in Boost Python 1.58.0 but it looks like 99022d2eb8fab7a433321b8b67249b87bacc44fb reverted it for 1.59.0 and the issue is present in all versions after. 

It looks as though it was changed back because of issues with Python 2.7, but the original problem with closing files still exists with Python 3.6.

Tested with 1.67.0 and Python 3.6.4.

Code used to reproduce:

```c++
#include <iostream>
#include "boost/python/detail/wrap_python.hpp"
#include "boost/python.hpp"

using namespace boost::python;

int main()
{
	int count = 0;
	Py_Initialize();

	while (count < 1000)
	{
		auto strScriptFile = "test.py";

		object main_module = import("__main__");
		dict globalDict = extract<dict>(main_module.attr("__dict__"));

		try
		{
			std::cout << "Run number: " << count << std::endl;
			exec_file(strScriptFile, globalDict);
		}
		catch (const error_already_set&)
		{
			std::cout << "Failed." << std::endl;
			break;
		}
		count++;
	}

	return 0;
}
```

hajokirchhoff:
This fixes #239 and #371

### DIFF
--- a/src/exec.cpp
+++ b/src/exec.cpp
@@ -129,6 +129,7 @@ object BOOST_PYTHON_DECL exec_file(char const *filename, object global, object l
                 f,
                 Py_file_input,
 		global.ptr(), local.ptr());
+  fclose(fs);
   if (!result) throw_error_already_set();
   return object(detail::new_reference(result));
 }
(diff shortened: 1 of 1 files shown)


hajokirchhoff:
This pull request should be pretty straightforward. It would be great, if these two very simple fixes could be included in the upcoming revision.

PR #373, opened 2021-08-15
Update call_method.hpp to avoid deprecation issue on PyEval

denisarnaud:
Was missing from https://github.com/boostorg/python/pull/320
I've tested it on one of my projects with (that patch on) Boost.Python/Boost 1.76.0 and it works well. Without that patch, there is a deprecation error.


hdu-sdlzx:
ping@stefanseefeld

stefanseefeld (approved):
Thanks !

PR #387, opened 2022-06-04
dynamic_cast before destructor

vitalybuka:
Call to the destructor ends lifetime of the object, including vptr
used by dynamic_cast.


vitalybuka:
ping

vitalybuka:
ping

PR #398, opened 2022-10-05
Support newer version of Sphinx

sdarwin:
Hi,
The current boostorg/boost bundle has been created using Sphinx version 1.5.6 

If upgrading _all_ packages and Sphinx changes to 5.2.1, the numpy docs are slightly broken. 

- Search doesn't work  
- bullet lists are widely spaced   

Researched the problems in https://github.com/sphinx-doc/sphinx/issues . After some experimentation it looks like the fix in this pull request solves the issues.

### DIFF
--- a/doc/numpy/_static/boost.css
+++ b/doc/numpy/_static/boost.css
@@ -714,3 +714,23 @@ span.purple { color: purple; }
 span.gold { color: gold; }
 span.silver { color: silver; } /* lighter gray */
 span.gray { color: #808080; } /* light gray */
+
+/* 2022 fix */
+
+ol.simple ol p,
+ol.simple ul p,
+ul.simple ol p,
+ul.simple ul p {
+    margin-top: 0;
+}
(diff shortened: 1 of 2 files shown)


sdarwin:
I tested these changes, and they worked for Sphinx 1.5.6 and 5.2.1. An easy way to verify is to merge this into the develop branch. New docs will appears at https://www.boost.org/doc/libs/develop/libs/python/doc/html/index.html .   If there are any problems it could be rolled back and not merged into the master branch.

Only a small number of boost libraries have any issues, such as this one. Most do not use Sphinx.  After this update, it will open the door to upgrade the entire set of packages for all boost docs.

sdarwin:
Sphinx 5.2.1 has been deployed in the boost develop branch. Not yet in the master branch. Example: https://www.boost.org/doc/libs/develop/libs/python/doc/html/numpy/reference/dtype.html

Issue #277, opened 2019-11-04
Support for generating pep484 .pyi stubs


chadrik:
[Static typing in python](https://realpython.com/python-type-checking/) (aka PEP-484) is gaining a lot of traction, with linting tools like [mypy](https://mypy.readthedocs.io/en/latest/) reaching maturity, and support in IDEs -- like PyCharm, VSCode, and Sublime -- for using pyi stubs to drive code inspection and completion.  Stubs are particularly impactful for c/c++ bindings which are opaque to static inspectors.  

pyi stub generation is already supported by other python binding generators like pybind11, shiboken, and sip.

Has this feature been discussed at all in boost-dev circles?

stefanseefeld:
I'm not aware of any such proposal, but I'd be open to consider any, notably if it has been tested with pybind11.

qhaas:
I started using a library built with `boost::python` recently and discovered nearly immediately that vscode and pyright were not detecting the types.  I had to frequently consult the source code of the library to determine typing information, which hampered productivity.

I forgot how reliant I have become on the features of PEP-484 and its standardized way of packaging typing information introduced in Python 3.7 with [PEP-561](https://peps.python.org/pep-0561/).  Including support for type hints would be a GREAT feature.

CEXT-Dan:
Started wrapping a very large C++ project, found it’s too hard to use without proper stubs.
So far I’ve been able to re-parse what boost::python emits to something semi usable, but it’s a combination of hand written and automation, very hard to maintain.  
I’m running embedded, couldn’t get routines like stubgen to work
Any improved support would be grateful

PR #417, opened 2023-05-11
Emit complete type information in generated docstrings

### LINKED ISSUE #426: __qualname__ on embedded/inner classes is wrong
Hi,

the __qualname__ on embedded classes and enum returns just the embedded class/enum class. The outer class is not included. Here is a simple example:

```
BOOST_PYTHON_MODULE(TestModule)
{
	class TestClass
	{
	public:
		enum class EmbeddedEnum
		{
			eOff,
			eRepeat,
			eMirror
		};

		class EmbeddedClass
		{
			EmbeddedClass() = default;
		};

		TestClass() = default;
	};

	boost::python::class_<TestClass, boost::shared_ptr<TestClass>, boost::noncopyable> boostTestClass("TestClass", boost::python::no_init);
	boost::python::scope boostTestClassScope = boostTestClass;

	boost::python::enum_<TestClass::EmbeddedEnum> boostTestClassEmbeddedEnum("EmbeddedEnum");
	boostTestClassEmbeddedEnum.value("off", TestClass::EmbeddedEnum::eOff);
	boostTestClassEmbeddedEnum.value("repeat", TestClass::EmbeddedEnum::eRepeat);
	boostTestClassEmbeddedEnum.value("mirror", TestClass::EmbeddedEnum::eMirror);
	boostTestClassEmbeddedEnum.export_values();

	boost::python::class_<TestClass::EmbeddedClass, boost::shared_ptr<TestClass::EmbeddedClass>, boost::noncopyable> boostEmbeddedTestClass("EmbeddedClass", boost::python::no_init);
}
```

`TestModule.TestClass.EmbeddedClass.__qualname__` returns just `EmbeddedClass`. Similarly the `TestModule.TestClass.EmbeddedEnum` returns `EmbeddedEnum`.

It should be `TestClass.EmbeddedClass` and `TestClass.EmbeddedEnum` respectively.

jvansanten:
PEP 484 type stubs are extremely convenient, both for static type checking with e.g. mypy and code completion with e.g. PyLance in VSCode or Jedi in vim. There are various ways to generate these for compiled extensions.

The `stubgen` that ships with mypy can generate stubs for compiled extensions, but knows nothing about types, so you would still have to complete them by hand. This is not exactly in the spirit of programatically-generated bindings. `pybind11-stubgen` does a similar thing, but parses the docstrings generated by `pybind11` to get argument and return types. While pybind11 can be [adapted to parse the signature format that boost-python already uses](https://github.com/jvansanten/pybind11-stubgen/tree/boostmode), the type information in the current generated docstrings is incomplete.

This PR introduces minimal changes such that the generated signatures can be used to create correct stubs. Specifically, these are:

- Set [`__qualname__`](https://peps.python.org/pep-3155/) on classes for Python >= 3.3 so that classes bound in `scope()` can be found again
- Use fully-qualified type names for imported types in signatures
- Emit signatures for property getters/setters
- Ensure that `slice` and `back_reference<T>` argument and return types are correctly annotated.

Existing tests were extended to verify the expected docstring contents.

This is a roundabout way of achieving #277, and seemed the most straightforward route given that pybind11 also seems to still rely on external tooling to generate stubs. Along the way, it also fixes #426.

Comments welcome!


stefanseefeld:
Hi @jvansanten , I unfortunately only now notice that with this PR merged some tests fail when run with Python 2. (See the error message "AttributeError: attribute '__doc__' of 'instancemethod' objects is not writable" in https://github.com/boostorg/python/actions/runs/10914059488/job/30291560557#step:5:1122) I don't yet fully understand the cause, i.e. whether the C++ implementation makes assumptions about the internal Python API, or whether it's only the test logic that needs to be conditionalized. Please let me know if you have any suggestions.
(There was also another minor glitch with the new `qualname()` function definition in `class.cpp` that returned a `str` rather than `object`, which caused an error on Windows. I have a tentative fix for that in https://github.com/boostorg/python/pull/451/commits/80593febbc1b3b2e18f8ba6d46a47fb44b380933.)

jvansanten:
Thanks for taking the time to merge this, @stefanseefeld, I really appreciate it!

Also, sorry for the breakage! I did not test this against Python2; that was sloppy of me. There are fixes in #452.

stefanseefeld:
Many thanks @jvansanten , I'll review and merge that shortly. (There are also some independent issues with the build logic, i.e. some of our test platforms are outdated and need some work. So not all failures are caused by issues in the code... :-) )

PR #422, opened 2023-08-30
Fix UB in instance_dealloc

### LINKED ISSUE #421: NULL ptr deref in instance_dealloc
Hi, we've recently triggered null pointer deref in [that place](https://github.com/boostorg/python/blob/47d5bc76f69e20625214381c930a2fad5765e2b3/src/object/class.cpp#L337C49-L337C71), right on dynamic_cast. From the first glance code does not look correct, dynamic_cast occurs after explicitly calling destructor on object. In our combination of clang16, gcc13 and python3.11 (I don't know what exactly triggered this) this destructor compiles in such way that vtable pointer is set to 0 in the end, which does not work so well with dynamic_cast.

Using object after destroying it sound like UB anyway, can it be fixed somehow? Using dynamic_cast before calling object destructor?

vient:
Do not use dynamic_cast after destroying object, it is not correct and leads to NULL pointer dereference with some compilers

This fix worked for me, no more SIGSEGVs.

Fix #421

### DIFF
--- a/src/object/class.cpp
+++ b/src/object/class.cpp
@@ -333,8 +333,9 @@ namespace objects
           for (instance_holder* p = kill_me->objects, *next; p != 0; p = next)
           {
               next = p->next();
+              void* topmost_object = dynamic_cast<void*>(p);
               p->~instance_holder();
-              instance_holder::deallocate(inst, dynamic_cast<void*>(p));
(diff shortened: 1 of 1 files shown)


stefanseefeld:
This was taken care of by #387

PR #425, opened 2023-11-13
Pass alignment to holder::allocate

vslashg:
aca3c8 added an alignment argument to holder::allocate which defaults to 1.  However, make_constructor.hpp uses this function to create storage for a pointer and was never changed to pass an alignment, so it uses the too-permissive default.

This causes unaligned pointers to be created on most architectures (although the x86 family tolerates it).  This can be detected as a failure by running projects under Clang ubsan.


stefanseefeld:
This fix is incorporated via #392

Issue #449, opened 2024-09-04
Heap overflow caused by incorrect alignment calculation in instance_holder::allocate()


andrew-rogers-wdc:
In the following code in [`libs\python\src\object\class.cpp`](https://github.com/boostorg/python/blob/8dd151177374dbf0aa5cb86bd350cf1ad13e2160/src/object/class.cpp#L750-L765):

```c++
typedef unsigned int alignment_marker_t;
                                                     
void* instance_holder::allocate(PyObject* self_, std::size_t holder_offset, std::size_t holder_size, std::size_t alignment)
{
    assert(PyType_IsSubtype(Py_TYPE(Py_TYPE(self_)), &class_metatype_object));
    objects::instance<>* self = (objects::instance<>*)self_;
    
    int total_size_needed = holder_offset + holder_size + alignment - 1;
    
    if (-Py_SIZE(self) >= total_size_needed)
    {
        // holder_offset should at least point into the variable-sized part
        assert(holder_offset >= offsetof(objects::instance<>,storage));

        size_t allocated = holder_size + alignment;
        void* storage = (char*)self + holder_offset;
        void* aligned_storage = ::boost::alignment::align(alignment, holder_size, storage, allocated);

        // Record the fact that the storage is occupied, noting where it starts
        const size_t offset = reinterpret_cast<uintptr_t>(aligned_storage) - reinterpret_cast<uintptr_t>(storage) + holder_offset;
        Py_SET_SIZE(self, offset);
        return (char*)self + offset;
    }
    else
    {
        const size_t base_allocation = sizeof(alignment_marker_t) + holder_size + alignment - 1;
        void* const base_storage = PyMem_Malloc(base_allocation);
       if (base_storage == 0)
            throw std::bad_alloc();

        const uintptr_t x = reinterpret_cast<uintptr_t>(base_storage) + sizeof(alignment_marker_t);
        //this has problems for x -> max(void *)
        //const size_t padding = alignment - ((x + sizeof(alignment_marker_t)) % alignment);
        //only works for alignments with alignments of powers of 2, but no edge conditions
        const uintptr_t padding = alignment == 1 ? 0 : ( alignment - (x & (alignment - 1)) );
        const size_t aligned_offset = sizeof(alignment_marker_t) + padding;
        void* const aligned_storage = (char *)base_storage + aligned_offset;
        BOOST_ASSERT((char *) aligned_storage + holder_size <= (char *)base_storage + base_allocation);
[...]

PR #450, opened 2024-09-04
Fix heap overflow

andrew-rogers-wdc:
* Quell size warning:
  ```
  libs\python\src\object\class.cpp(732): warning C4267: 'initializing': conversion from 'size_t' to 'int', possible loss of data
  ```
* Fix heap overflow (#449).


andrew-rogers-wdc:
This fix was incorporated via #392.

PR #463, opened 2024-12-14
Prevent potential data loss in numpy's dtype::get_itemsize().

JohannesWilde:
The numpy 2.0 migration guide states [https://numpy.org/devdocs/numpy_2_0_migration_guide.html]:

"PyDataType_ELSIZE and PyDataType_SET_ELSIZE (note that the result is now npy_intp and not int)."

Even though it is rather unlikely to have have elements larger than 2GB each, prevent potential numerical overflows, which could occur when casting ssize_t to int on 64-bit systems.

### DIFF
--- a/include/boost/python/numpy/dtype.hpp
+++ b/include/boost/python/numpy/dtype.hpp
@@ -47,7 +47,7 @@ public:
   template <typename T> static dtype get_builtin();
 
   /// @brief Return the size of the data type in bytes.
-  int get_itemsize() const;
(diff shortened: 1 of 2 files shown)


JohannesWilde:
As before, there is no check for numerical overflows in any of the multplications in `is_c_contiguous` and `is_f_contiguous` when doing `total *= (*i);`. Shoudl someone change this?

JohannesWilde:
The file `ndarray.cpp` still contains several tab characters. Do you have coding rules relating to tabs?

PR #465, opened 2024-12-21
Replace Boost::foreach dependency with range-based for

georgthegreat:
While the library is still marked as c++03 compatible I wonder if the time has come to follow the rest of the boost libraries and raise it up to c++11.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -63,7 +63,6 @@ target_link_libraries(${_boost_python}
     Boost::conversion
     Boost::core
     Boost::detail
-    Boost::foreach
     Boost::function
     Boost::iterator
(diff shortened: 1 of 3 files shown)


Osyotr:
These public dependencies require C++11:
- Boost::align
- Boost::bind
- Boost::conversion
- Boost::function
- Boost::iterator
- Boost::lexical_cast
- Boost::smart_ptr

It is perfectly reasonable to drop C++03 support in Boost::python, too.

e-kwsm (commented):
https://github.com/boostorg/python/blob/4fc3afa3ac1a1edb61a92fccd31d305ba38213f8/.github/workflows/test-ubuntu.yml#L14

drop c++98?

PR #470, opened 2025-01-28
Replace use of boost/iterator/detail/enable_if.hpp

pdimov:
`boost/iterator/detail/enable_if.hpp` has been deleted. Replace with `boost/type_traits/enable_if.hpp`.

### DIFF
--- a/include/boost/python/object_operators.hpp
+++ b/include/boost/python/object_operators.hpp
@@ -9,7 +9,7 @@
 
 # include <boost/python/object_core.hpp>
 # include <boost/python/call.hpp>
-# include <boost/iterator/detail/enable_if.hpp>
+# include <boost/type_traits/enable_if.hpp>
 # include <boost/mpl/bool.hpp>
 
 # include <boost/iterator/detail/config_def.hpp>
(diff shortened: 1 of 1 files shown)


pdimov:
Note that Boost.Python currently fails to build without this PR, so it should be applied sooner rather than later, and preferably merged to master as well.

OgreTransporter:
https://github.com/boostorg/python/blob/4fc3afa3ac1a1edb61a92fccd31d305ba38213f8/include/boost/python/object_operators.hpp#L43

Or use `std::enable_if` instead of `boost::iterators::enable_if`.

pdimov:
I'm merging this to unblock the [Boost CI](https://github.com/boostorg/boost/actions/workflows/ci.yml) which fails because Boost.Python fails to build.

francesco-ballarin:
@pdimov can this be backported to `master`? The goal would be for the update to go through to the `master` branch at https://github.com/boostorg/boost.

Thanks,
Francesco

pdimov:
@stefanseefeld what's your merge to master policy? I don't want to do this without your approval.

stefanseefeld:
@pdimov thanks for asking, let me take care of it right away...done.

PR #472, opened 2025-02-01
Fix homebrew include path.

stefanseefeld:


pdimov:
This seems to have worked, but the build now fails for a different reason, Optional requires C++11. I see that the command line passes `cxxflags=-std=c++11`, but then the log says
```
configuration check results:
  has_cxx11 : False 
  has_numpy : False 
```
so it looks like this doesn't take for some reason.

Maybe `faber` should be modified to print the command line on failure, like `b2` does.

The problem could be that the `makedep` rule doesn't apply `cxxflags`.

stefanseefeld:
Good catch @pdimov . Indeed, `faber` users the `cppflags` feature to configure the pre-processor, and `cxxflags` for the (C++) compiler. Setting the `-std` option in both has fixed the build problem. Thanks !

Issue #479, opened 2025-03-28
test/upcast.cpp does not compile since python 3.10


e-kwsm:
revision: 3e7be69e1e405e1d5ddd232c69c024ee441592c5

Since Python 3.10, compilation of test/upcast.cpp fails.

https://github.com/boostorg/python/blob/3e7be69e1e405e1d5ddd232c69c024ee441592c5/test/upcast.cpp#L16-L17

```console
$ g++ -Iinclude -I/opt/miniconda3/envs/py310/include/python3.10 test/upcast.cpp
In file included from /usr/include/boost/detail/lightweight_test.hpp:15,
                 from ./test/upcast.cpp:7:
./test/upcast.cpp: In function 'int main()':
/opt/miniconda3/envs/py310/include/python3.10/object.h:133:33: error: lvalue required as unary '&' operand
  133 | #define Py_REFCNT(ob) _Py_REFCNT(_PyObject_CAST_CONST(ob))
      |                       ~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~
./test/upcast.cpp:16:17: note: in expansion of macro 'Py_REFCNT'
   16 |     BOOST_TEST(&Py_REFCNT(boost::python::upcast<PyObject>(&o)) == &Py_REFCNT(&o));
      |                 ^~~~~~~~~
/opt/miniconda3/envs/py310/include/python3.10/object.h:133:33: error: lvalue required as unary '&' operand
  133 | #define Py_REFCNT(ob) _Py_REFCNT(_PyObject_CAST_CONST(ob))
      |                       ~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~
./test/upcast.cpp:16:68: note: in expansion of macro 'Py_REFCNT'
   16 |     BOOST_TEST(&Py_REFCNT(boost::python::upcast<PyObject>(&o)) == &Py_REFCNT(&o));
      |                                                                    ^~~~~~~~~
/opt/miniconda3/envs/py310/include/python3.10/object.h:133:33: error: lvalue required as unary '&' operand
  133 | #define Py_REFCNT(ob) _Py_REFCNT(_PyObject_CAST_CONST(ob))
      |                       ~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~
./test/upcast.cpp:17:17: note: in expansion of macro 'Py_REFCNT'
[...]

PR #502, opened 2025-10-29
Updated to recent compilers/boost

teeks99:


e-kwsm:
It seems that Boost.Function requires C++11 since 1.85 (https://github.com/boostorg/function/commit/f86ddb0bb977946c64ba22b8456fb6551c6d04ac), and upcast fails since python is now 3.12 (#479).

teeks99:
I finally got a decent gcc-15 build, that has been pushed as well, it looks like everything (except the C++98 jobs) is getting past build and into test.

PR #509, opened 2026-01-22
Remove dependencies on Boost.StaticAssert

Lastique:
Boost.StaticAssert has been merged into Boost.Config, so remove the dependency.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -71,7 +71,6 @@ target_link_libraries(${_boost_python}
     Boost::numeric_conversion
     Boost::preprocessor
     Boost::smart_ptr
-    Boost::static_assert
     Boost::tuple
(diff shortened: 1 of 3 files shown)


pdimov:
I'm going to merge this, because without it we get errors like
```
error: Unable to find file or target named
error:     '/boost/static_assert//boost_static_assert'
error: referred to from project at
error:     '/home/runner/work/boost_install/boost-root/libs/python/build'
error: could not resolve project reference '/boost/static_assert'
```
now.

pdimov:
This PR needs to go into `master` as well, but since there are nontrivial commits on `develop`, I won't merge it. Please take care of that.

stefanseefeld:
I will have a look in the next couple of days. Thanks.

PR #518, opened 2026-06-04
Try to fix Windows CI (GitHub Actions -- vcpkg manifest mode)

### LINKED ISSUE #514: github actions are outdated
https://github.com/boostorg/python/blob/32da86df269fb8f7ef777ddd8c3424d854e94455/.github/workflows/test-windows.yml#L20

The above gives

https://github.com/boostorg/python/actions/runs/23999475219/job/69993078166?pr=513#step:5:28

> Run lukka/run-vcpkg@v6
> Warning: The `save-state` command is deprecated and will be disabled soon. Please upgrade to using Environment Files. For more information see: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

This is already fixed in upstream (https://github.com/lukka/run-vcpkg/issues/168); the latest version is v11&mdash;I&rsquo;m not sure why v8 and v9 are skipped though.

stephengtuggy:
Fixes #514 (hopefully)

### DIFF
--- a/.github/workflows/test-windows.yml
+++ b/.github/workflows/test-windows.yml
@@ -72,12 +72,12 @@ jobs:
     - name: build
       shell: cmd
       run: |
-        faber --builddir=build cxx.name=msvc --log=commands --log=output --with-boost-include="${{ github.workspace }}/vcpkgInstalled" -j4
+        faber --builddir=build cxx.name=msvc --log=commands --log=output --with-boost-include="${{ github.workspace }}/vcpkgInstalled/x64-windows/include" -j4
 
     - name: test
       shell: cmd
       run: |
-        faber --builddir=build cxx.name=msvc --with-boost-include="${{ github.workspace }}/vcpkgInstalled" -j4 test.report
+        faber --builddir=build cxx.name=msvc --with-boost-include="${{ github.workspace }}/vcpkgInstalled/x64-windows/include" -j4 test.report
 
     - name: Save vcpkg cache
       id: cache-vcpkg-save
--- a/.github/workflows/test-windows.yml
+++ b/.github/workflows/test-windows.yml
@@ -16,6 +16,7 @@ jobs:
       X_VCPKG_ASSET_SOURCES: clear;x-azurl,file:///${{ github.workspace }}/vcpkgAssets,,readwrite
       VCPKG_BINARY_SOURCES: clear;files,${{ github.workspace }}/vcpkgBin,readwrite
       VCPKG_BUILD_TREES: "${{ github.workspace }}/vcpkgBuild"
+      VCPKG_INSTALLED_DIR: "${{ github.workspace }}/vcpkgInstalled"
       VCPKG_INSTALL_OPTIONS: "--x-buildtrees-root=${{ github.workspace }}/vcpkgBuild"
       VCPKG_DEFAULT_TRIPLET: 'x64-windows'
       VCPKG_DEFAULT_HOST_TRIPLET: 'x64-windows'
@@ -71,12 +72,12 @@ jobs:
     - name: build
       shell: cmd
       run: |
-        faber --builddir=build cxx.name=msvc --log=commands --log=output --with-boost-include="${{ github.workspace }}\\969f6665-88a2-4c98-938e-ca9259871fec\\vcpkg_installed\\x64-windows\\include" -j4
+        faber --builddir=build cxx.name=msvc --log=commands --log=output --with-boost-include="${{ github.workspace }}/vcpkgInstalled" -j4
 
     - name: test
       shell: cmd
       run: |
-        faber --builddir=build cxx.name=msvc --with-boost-include="${{ github.workspace }}\\969f6665-88a2-4c98-938e-ca9259871fec\\vcpkg_installed\\x64-windows\\include" -j4 test.report
+        faber --builddir=build cxx.name=msvc --with-boost-include="${{ github.workspace }}/vcpkgInstalled" -j4 test.report
 
     - name: Save vcpkg cache
       id: cache-vcpkg-save
--- a/.github/workflows/test-windows.yml
+++ b/.github/workflows/test-windows.yml
@@ -25,6 +25,9 @@ jobs:
       with:
         vcpkgDirectory: '${{ github.workspace }}/vcpkg'
         runVcpkgInstall: true
+    - name: List directory contents
+      run: Get-ChildItem "${{ github.workspace }}/vcpkg" -Recurse -Force -File -Filter 'config.hpp'
+      shell: pwsh
     - name: setup faber
       run: |
         python -m pip install --upgrade pip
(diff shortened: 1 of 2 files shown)


stefanseefeld:
thanks @stephengtuggy for the patch ! It's definitely a step in the right direction, but apparently not yet complete (the build fails).

stephengtuggy:
> thanks @stephengtuggy for the patch ! It's definitely a step in the right direction, but apparently not yet complete (the build fails).

@stefanseefeld you're welcome. I think I see what the problem is. I neglected to update the include directory passed to the `faber` commands. That should be fixed now.

stephengtuggy:
Nope, still not right. And unfortunately, this may not be trivial to fix. I'm not sure if there is a singular Boost include directory that can be passed to faber. I think it may be more like a different include directory for each Boost component. How to handle this situation?

stephengtuggy:
To eliminate a variable, I am going back to the same vcpkg baseline in the manifest file as was used for `vcpkgGitCommitId` prior to this PR. Let's see if that helps.

.github/workflows/test-windows.yml
```diff
@@ -10,30 +10,24 @@ jobs:
...
-          boost-lexical-cast
-          boost-mpl
-          boost-preprocessor
-          boost-smart-ptr
-          boost-static-assert
-          boost-align
+        vcpkgDirectory: '${{ github.workspace }}/vcpkg'
+        runVcpkgInstall: true
+    - name: List directory contents
+      run: Get-ChildItem "${{ github.workspace }}/vcpkg" -Recurse -Force -File -Filter 'config.hpp'
```
stefanseefeld: Perhaps if you use `-User .` to start looking in the local directory instead of $github.workspace/vcpkg ?
Alternatively, I see that the install root can be set via the `vcpkgArguments: '--x-manifest-root=${{ github.workspace }}/vcpkg'` directive.

stephengtuggy: This command simply searches the `<github workspace>/vcpkg` directory and all subdirectories for a file named config.hpp. It is literally the equivalent of `find "${{ github.workspace}}/vcpkg" -iname 'config.hpp' -type f`. And it's not finding the file anywhere! I don't understand what is going on here. Why would this file not be present?

I'm not familiar with `-User .` in conjunction with the Get-ChildItem command. What would that do?

stephengtuggy: Yes, the install root can be set using the `vcpkgArguments` parameter, but I'm not sure it's best practice to do so.

stephengtuggy:
It occurs to me that in my other projects that use vcpkg, I generally set the `VCPKG_ROOT` environment variable explicitly. This might be important. Trying it now.

stephengtuggy:
Well, I can finally see where config.hpp is. It's in directory `D:\a\python\python\969f6665-88a2-4c98-938e-ca9259871fec\vcpkg_installed\x64-windows\include\boost`. In other words, `'${{ github.workspace }}\<GUID>\vcpkg_installed\$VCPKG_DEFAULT_TRIPLET\include'` would be the boost include directory setting for faber.

This raises some more questions: why? Where is this setting coming from? And does the GUID stay the same, or does it change every run?

stefanseefeld:
That's great progress ! But I agree, having to remember and hardcode such paths is cumbersome and fragile, so best to be avoided. So why not replace all that by a manually defined install root (specified vie `vcpkgArguments`) ?

stephengtuggy:
> That's great progress ! But I agree, having to remember and hardcode such paths is cumbersome and fragile, so best to be avoided. So why not replace all that by a manually defined install root (specified vie `vcpkgArguments`) ?

Yes, I guess that makes sense, actually. I was trying to set it using environment variables, but that doesn't seem to be working for some reason. Perhaps something closer to where vcpkg runs is overriding the relevant environment variable.

At any rate, the last run worked! So the GUID seems to stay the same with each run.

stephengtuggy:
I don't think `--x-manifest-root` is the correct vcpkg argument for this purpose, by the way. I believe `--x-manifest-root` tells vcpkg where to find the manifest file (vcpkg.json), not where to install the packages it installs.

stephengtuggy:
[Looks like](https://github.com/microsoft/vcpkg/issues/22279#issuecomment-1005999694) one can tell vcpkg where to install dependencies using the `VCPKG_INSTALLED_DIR` environment variable. I will try that.

.github/workflows/test-windows.yml
```diff
@@ -10,40 +10,83 @@ jobs:
...
+      VCPKG_BINARY_SOURCES: clear;files,${{ github.workspace }}/vcpkgBin,readwrite
+      VCPKG_BUILD_TREES: "${{ github.workspace }}/vcpkgBuild"
+      VCPKG_INSTALLED_DIR: "${{ github.workspace }}/vcpkgInstalled"
+      VCPKG_INSTALL_OPTIONS: "--x-buildtrees-root=${{ github.workspace }}/vcpkgBuild"
+      VCPKG_DEFAULT_TRIPLET: 'x64-windows'
+      VCPKG_DEFAULT_HOST_TRIPLET: 'x64-windows'
+    
     steps:
-    - uses: actions/checkout@v5
+    - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7
```
stefanseefeld, line 25: I see you are referring to all actions by sha. Is that really necessary ? Why not use a version tag instead ? Wouldn't that be simpler (more readable, easier to maintain, etc.) ?

stephengtuggy: Referring to GitHub Actions by SHA is considered a security best practice. See <https://docs.github.com/en/actions/reference/security/secure-use?learn=getting_started#using-third-party-actions> .

It's not too difficult to maintain if one uses a tool like Dependabot or Renovate[bot] to keep dependencies up to date, including GitHub Actions dependencies.

stephengtuggy: As for readability, I do at least include the version tag in a comment. Hopefully that helps at least a little bit.

stefanseefeld: Yes, I noticed. Thanks !
Please go ahead and merge this if you are satisfied with the PR. Many thanks again for all your effort on this ! I wouldn't have been able to pull this off on my own, as I haven't developed on & for Windows in a long time.
Your work will unblock a number of other changes that I was hesitant to add while the Windows build was broken, so I'm thrilled to see this being completed.
Cheers !

stephengtuggy: You're welcome! I am satisfied with the PR, but I don't see a Merge button. I probably don't have merge permissions at the moment.

stefanseefeld: Ah, makes sense. I suppose only team members have merge (or commit) permission. Let me merge this for now. If you want to contribute in the future, we can consider adding you to the team. Cheers !

PR #522, opened 2026-06-27
Windows CI: Exit workflow on any errors; fix vcpkg caching

### LINKED ISSUE #521: Windows CI does not fail if one or more tests fail
While I was working on updating the Windows CI, I noticed that some tests were failing, but the test suite -- and the GitHub Actions workflow as a whole -- succeeded anyway. We should probably change that.

### LINKED ISSUE #523: Windows CI: vcpkg dependency caching not working

stephengtuggy:
Fixes #521 (hopefully).

Fixes #523 .


stefanseefeld:
Can you explain the change ? It looks to me like you merely replaced `cmd` by `pwsh` and changed the commands accordingly. How does that affect the caching ?

stephengtuggy:
> Can you explain the change ? It looks to me like you merely replaced `cmd` by `pwsh` and changed the commands accordingly. How does that affect the caching ?

Yes, I can explain. The cache save step was failing with an error that the paths given did not exist -- even though they did. Removing the quotes from the paths fixed that. The cache save step is now succeeding, and at least some of the time, the cache restore step is now also succeeding, because there is a cache entry to retrieve.

I also fixed the cache key calculation, which was incorrect. I originally copy-pasted it from one of my other repositories, and it included some matrix parameters that are not valid in this repository. That is now fixed.

The other changes may or may not be relevant. As you said, I changed the shell used to PowerShell, and changed the commands accordingly. I have a little bit more confidence that PowerShell will actually fail the build if a test fails, vs. cmd.exe. But I'm not 100% sure that it really makes a difference.

stefanseefeld:
OK, but the two shell changes (from `cmd` to `pwsh` in the `build` and `test` steps) are unrelated and not needed, yes ?

stephengtuggy:
> OK, but the two shell changes (from `cmd` to `pwsh` in the `build` and `test` steps) are unrelated and not needed, yes ?

From [this SuperUser question](https://superuser.com/questions/1482950/make-cmd-exe-exit-on-any-error-in-a-batch-file) and [this StackOverflow one](https://stackoverflow.com/q/9948517/5067822), it seems that switching to PowerShell is necessary in order to stop on any error, but by itself it's not sufficient. I will need to [add some more logic](https://stackoverflow.com/a/9949105/5067822).

stephengtuggy:
> > OK, but the two shell changes (from `cmd` to `pwsh` in the `build` and `test` steps) are unrelated and not needed, yes ?
> 
> From [this SuperUser question](https://superuser.com/questions/1482950/make-cmd-exe-exit-on-any-error-in-a-batch-file) and [this StackOverflow one](https://stackoverflow.com/q/9948517/5067822), it seems that switching to PowerShell is necessary in order to stop on any error, but by itself it's not sufficient. I will need to [add some more logic](https://stackoverflow.com/a/9949105/5067822).

Done. @stefanseefeld please review

PR #524, opened 2026-07-01
Add test for multi-threaded import of multi-phase module

o01eg:
Also tests module state isolation

### DIFF
--- a/test/multi_thread_import_.cpp
+++ b/test/multi_thread_import_.cpp
@@ -1,0 +1,165 @@
+// Copyright Stefan Seefeld 2007.
+// Distributed under the Boost Software License, Version 1.0. (See
+// accompanying file LICENSE_1_0.txt or copy at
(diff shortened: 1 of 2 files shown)


test/multi_thread_import_.cpp
```diff
@@ -0,0 +1,165 @@
...
+  for (auto& t : threads)
+  {
+    t.join();
+  }
+
+  // Restore the main thread's GIL
+  PyEval_RestoreThread(main_thread_state);
+    
+  // Boost.Python doesn't support Py_Finalize yet.
+  // Py_Finalize();
```
Osyotr, line 157: Is this actually true?

o01eg: Same line is still in `test/import_.cpp`.

Osyotr: It was authored before git migration and that was in 2007...

